PR #2, opened 2020-03-16
new API calls 

ethansiegl:
- move getContent() to util file to reuse in other components, 
- add tsconfig file
- add wordpress API calls to populate top level tab content
- add next.env file


tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+  "compilerOptions": {
+    "target": "es5",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": false,
```
onursumer: I thought I have included a tsconfig file in my PR, I might have forgot to add that. Can we set strict to true and see if anything fails?

util.tsx
```diff
@@ -0,0 +1,15 @@
+import useSWR from "swr";
```
onursumer: Does this file need to be a `tsx`? As long it doesn't return any JSX content, it is better to name it as `util.ts`. Also, it might be better to have something more specific than a generic name like  `util.ts`.

ethansiegl: Sure I can rename the extension. I'll also rename the filename to `api_util.ts`

tsconfig.json
ethansiegl: yes the file was not committed. I can set to true, build seems to be ok

util.tsx
onursumer: Thanks @ethansiegl. Can we use camel case (`apiUtil.ts`) or title case (`ApiUtil.ts`) instead? I prefer title case in general, but of course there is no universal rule for file naming. We shouldn't use title case for files under `pages` though, because they're mapped to actual routes.

PR #70, opened 2020-11-03
Add static chart with numbers

inodb:
Add static charts with data currently being collected on homepage (numbers not final).
Remove cards temporarily.

Screen Shot 2020-11-03 at 9 16 02 AM


inodb:
Decided to leave this out b/c we don't want to show potential discrepancies with actual ingress status

PR #75, opened 2021-01-20
try to import synapse-react-client

inodb:
currently errors with global CSS cannot be imported from within
node_modules

### DIFF
--- a/components/filter/filterSelection.module.scss
+++ b/components/filter/filterSelection.module.scss
@@ -11,7 +11,7 @@
     position: absolute;
     left: 0;
     height: 100px;
-    width: 96vw;
+    /*width: 96vw;*/
     background: #fff;
     border: 1px solid black;
 }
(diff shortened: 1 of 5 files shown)


onursumer:
To fix the mismatching react versions the workaround is to add `resolutions` into `package.json`. Ideally the library should make `react` a peer dependency instead of a dependency.

```JavaScript
  "resolutions": {
    "synapse-react-client/react": "16.13.0"
  }
```

Some components have direct reference to the `window` object which causes issues for server side rendering. So, for example for the `Login` component, we need to disable server side rendering.

```JavaScript
const Login = dynamic(
    () => import('synapse-react-client/dist/containers/Login'),
    { ssr: false }
);
```

Not sure how to fix the global css reference issue but ideally the library itself should not have any reference to external global css files. Consumers of the library should include them explicitly.

inodb:
@milen-sage Could you help loop in someone who can help address @onursumer's comments? Thanks!

inodb:
It would be ideal if we can pull the package from npm as well (as im doing currently in this PR)

inodb:
Closing this for now

Issue #77, opened 2021-01-25
Data Release Table Improvements


inodb:
on https://data.humantumoratlas.org/explore

- [x] Add loading indicator (Done in https://github.com/ncihtan/htan-portal/pull/78)

- [x] Add nicer loading indicator (ADAM) (Done in https://github.com/ncihtan/htan-portal/pull/93)

- [x] Add route/parameter for tabs (default listing can be `/explore?tab=atlas` then when you go to files tab it is `/explore?tab=files` (ADAM) (Done in https://github.com/ncihtan/htan-portal/pull/92)

- [x] Add file type filter button and tooltip: (ADAM) (Done in https://github.com/ncihtan/htan-portal/pull/106)

    <img width="250" alt="Screen Shot 2021-01-25 at 4 18 54 PM" src="https://user-images.githubusercontent.com/1334004/105767444-0d843c80-5f29-11eb-9d91-5eee70036f50.png">

~- [ ] Add assay type: (ADAM)~ (Putting off more for now)

![image](https://user-images.githubusercontent.com/1334004/105768130-eda14880-5f29-11eb-9062-3b6c3ae2714d.png)

- [x] Assay Type First Iteration Improvements (https://github.com/ncihtan/htan-portal/issues/89)
 

- [x] Add Files as separate column aligned in tooltip as in above images (instead of showing it in parentheses with e.g. `Cancer Type (xxx)`)

- [x] Add filtering string to URL to allow for URL sharing (https://github.com/ncihtan/htan-portal/issues/96)

- [x] The download button should give explanation of how to download the selected files using the synapse command line client (#87)

- [x] Add BioSpecimen Tab (?)

- [x] Add Cases Tab (?)

- [x] Default should be atlas Tab (not files) (https://github.com/ncihtan/htan-portal/pull/111)
    - [x] Atlas Tab should list Atlas ID (https://github.com/ncihtan/htan-portal/pull/111)

- [x] Search filter should cover some more use cases:
   - [ ] ~search for single case id on case id~
   - [ ] ~search for fields on atlas page~
   - [x] search by biospecimen id on files tab
  
- [x] Add reset filters button (delete all filters) (Done in https://github.com/ncihtan/htan-portal/pull/110)
~cancelled - [x] purple download button remains blue after clicking on it (Done in https://github.com/ncihtan/htan-portal/pull/112)~

Bugs:

- [x] All filter search bug (https://github.com/ncihtan/htan-portal/issues/94)
[...]

PR #78, opened 2021-01-25
Add loading indicator while data is fetched in explore page

adamabeshouse:
Only uses placeholder for now.

Related to #77


inodb (approved):
place holder looks good to me - im fine with reusing cbioportal loading icon but anything works :)

Issue #89, opened 2021-02-02
Assay Type Filter Changes


inodb:
Could we change the assay type filter to just list the first part of the assay name, so this one:

<img width="281" alt="Screen Shot 2021-02-01 at 2 41 37 PM" src="https://user-images.githubusercontent.com/1334004/106509524-bbe53000-649b-11eb-97e5-f2dc8d438af2.png">

could be just five items:

```
scRNA-Seq
Bulk RNA-Seq
scATAC-Seq
Bulk WES
Other Assay
```

The levels are already captured by the File Type dropdown.

PR #90, opened 2021-02-04
Dont stratify assay types by level, since level can be filtered in another menu

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/89

### DIFF
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -102,11 +102,12 @@ export async function loadData(WPAtlasData: WPAtlas[]):Promise<LoadDataResult> {
 
     _.forEach(files, (file) => {
 
-        // make a new level property by parsing component
+        // parse component to make a new level property and adjust component property
         if (file.Component) {
-            const level = file.Component.match(/Level(\d+)/i);
-            if (level && level.length > 1) {
-                file.level = `Level ${level[1]}`;
+            const parsed = parseRawAssayType(file.Component);
+            file.Component = parsed.name;
+            if (parsed.level && parsed.level.length > 1) {
+                file.level = parsed.level;
             } else {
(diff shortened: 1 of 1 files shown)


lib/helpers.ts
```diff
@@ -168,4 +169,28 @@ export function sortStageOptions(options:ExploreOptionType[]){
...
+    const splitByLevel = t.split("Level");
+    const level = splitByLevel.length > 1 ? `Level ${splitByLevel[1]}` : null;
+    const extractedName = splitByLevel[0].match(/bts:(.+)/);
+    if (extractedName && extractedName[1]) {
+        // Convert camel case to space case
+        // Source: https://stackoverflow.com/a/15370765
+        let name = extractedName[1].replace( /([A-Z])([A-Z])([a-z])|([a-z])([A-Z])/g, '$1$4 $2$3$5' );
+
+        // Handle special case: scRNA
+        name = name.replace(/Sc RNA\b/g, "scRNA");
```
inodb, line 189: prolly anything prefixed with `Sc ` should be changed in this way 

e.g. `Sc ATAC-seq` should be `scATAC-seq`

inodb: sc is short for single cell

inodb (approved):
Thanks! LGTM - one minor comment about `Sc ATAC-Seq`

lib/helpers.ts
inodb: same applies to `sn` (single nucleus), see data standards list here:

https://htan-portal-nextjs-ln46lfjfo.vercel.app/standards

adamabeshouse: okay cool, Ill just always treat a lone sc as a prefix

PR #93, opened 2021-02-05
Prettier loading indicator

adamabeshouse:
We can't use react-spinkit because of this problem https://github.com/vercel/next-plugins/issues/129 (old Github issue but the problem still exists)

image


inodb:
@adamabeshouse Nice ✨ ! I like it

Minor thing: maybe nice to hide the footer during loading?

Screen Shot 2021-02-06 at 5 17 12 PM

inodb (approved):

PR #99, opened 2021-02-10
Enable type checking in dev mode

onursumer:
See https://github.com/vercel/next.js/issues/12735#issuecomment-629404102


inodb:
@onursumer did you see this comment: https://github.com/vercel/next.js/issues/12735#issuecomment-629404842. Do you experience a slowdown?

onursumer:
@inodb right now the slowdown is negligible since the codebase is not that large. We may need to come up with a different solution at some point when the build process gets significantly slower.

adamabeshouse:
RIP FastRefresh 💀

adamabeshouse (approved):

PR #101, opened 2021-02-10
Use the same table component for all explore tabs

### LINKED ISSUE #95: Use same table component for case and sample tabs on explore tab as for files

onursumer:
Fix #95


lib/dataTable.ts
```diff
@@ -0,0 +1,15 @@
+export function getDefaultDataTableStyle() {
```
adamabeshouse, line 1: `DataTableHelpers.ts`? `dataTableHelpers`?

adamabeshouse (approved):
looks good, just maybe a file name change

PR #103, opened 2021-02-17
Refactor explore page into multiple components

onursumer:


inodb (approved):
LGTM 🎉

components/filter/Filter.tsx
```diff
@@ -0,0 +1,113 @@
...
+                                        PropNames[
+                                            filter as keyof typeof PropNames
+                                            ]
+                                        ].displayName
+                                }
+                            </span>
+
+                            {props.selectedFiltersByGroupName[filter].map(
+                                (value, i, values) => {
+                                    const numberOfValues =
```
adamabeshouse, line 47: I think it would be good to move the logic computing the parens etc into a helper function - would make the body of `Filter` easier to read

adamabeshouse (approved):
Nice!

Issue #82, opened 2021-01-26
Home Page


inodb:
- [x] Cards of all atlases' should be filled in
- [x] Link cards to atlas pages
- [x] Make the cards look prettier
- [ ] Dynamically compute some of the stats shown on the homepage (@adamabeshouse )

PR #108, opened 2021-02-22
Dynamically calculate homepage stats based on data

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/82

### DIFF
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -68,6 +68,7 @@ export interface Entity {
     fileFormat: string;
     filename: string;
     HTANParentID: string;
+    HTANParticipantID:string;
     level: string;
     WPAtlas: WPAtlas;
     biospecimen: Entity | undefined;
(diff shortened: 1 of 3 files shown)


lib/helpers.ts
```diff
@@ -236,3 +238,22 @@ export function setTab(
         query: { tab }
     }, undefined, { shallow: true });
 }
+
+export function computeDashboardData(files:Entity[]) {
+    let uniqueAtlases = new Set();
+    let uniqueOrgans = new Set();
+    let uniqueBiospecs = new Set();
+    let uniqueCases = new Set();
```
onursumer, line 246: make these `const` instead?

onursumer (approved):
Looks good to me

lib/helpers.ts
adamabeshouse: oops, yeah

Issue #98, opened 2021-02-10
All filter search bug (can't deselect)


inodb:
I think it would make sense to remove the filter on click:

![htan_filter_deselect_bug](https://user-images.githubusercontent.com/1334004/107529566-fda36400-6b88-11eb-8070-048e5621404e.gif)

https://data.humantumoratlas.org/explore?selectedFilters=%5B%7B%22value%22%3A%22Pre-Cancer+Atlas+Pilot+Project+%28PCAPP%29%22%2C%22label%22%3A%22Pre-Cancer+Atlas+Pilot+Project+%28PCAPP%29%22%2C%22group%22%3A%22AtlasName%22%2C%22count%22%3A120%7D%2C%7B%22value%22%3A%22Human+Tumor+Atlas+Pilot+Project+%28HTAPP%29%22%2C%22label%22%3A%22Human+Tumor+Atlas+Pilot+Project+%28HTAPP%29%22%2C%22group%22%3A%22AtlasName%22%2C%22count%22%3A120%7D%5D

Expected behavior: when I click on the filter again it should deselect. Instead I currently can't deselect them

PR #110, opened 2021-02-25
Fix issues with filtering

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/98
Fixes bug " Removing filters by clicking on the group name doesn't work"
Adds button to clear all filters

### DIFF
--- a/pages/explore.tsx
+++ b/pages/explore.tsx
@@ -19,6 +19,7 @@ import {
     updateSelectedFiltersInURL
 } from '../lib/helpers';
 import {
+    ExploreActionMeta,
     ExploreSelectedFilter, FilterAction,
     IFilterProps,
     IFiltersByGroupName,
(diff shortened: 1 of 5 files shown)


pages/explore.tsx
```diff
@@ -84,24 +84,31 @@ class Search extends React.Component<{ router: NextRouter, wpData: WPAtlas[] },
...
-            // first remove the item
-            newFilters = this.selectedFilters.filter((o)=>{
-                return o.group !== actionMeta!.option!.group! || o.value !== actionMeta!.option!.value!;
-            });
-
-            if (actionMeta.action === 'select-option') {
-                const option = actionMeta.option;
-                newFilters = newFilters.concat([option]);
+            switch (actionMeta.action) {
+                case 'clear':
```
onursumer, line 93: should we define an enum for these actions? `FilterAction.Clear` or something like that

adamabeshouse: good point. done

alisman (approved):

PR #112, opened 2021-02-26
Defocus download button after clicking

adamabeshouse:


inodb (approved):

onursumer:
@adamabeshouse Do you know why the build is failing?

components/FileTable.tsx
```diff
@@ -209,8 +209,10 @@ export default class FileTable extends React.Component<IFileTableProps> {
         makeObservable(this);
     }
 
-    @action onDownload = () => {
+    @action onDownload = (evt:any) => {
         this.isDownloadModalOpen = true;
+        // Defocus "Download" button
+        evt.target.blur();
```
alisman, line 215: what does this fix?

adamabeshouse: its a UI request from niki

alisman:
I think this is kind of silly.  It's the default behavior of buttons in browser and has something to do with accessibility.  I would be in favor of not doing this

Issue #115, opened 2021-03-03
On Atlas Page Remove Tabs that already exist on Explore Page


inodb:
On this page: https://data.humantumoratlas.org/atlas/hta2. We were previously showing files. We have decided to only list the files on the explore page instead. So when one clicks on the "View Files" button it takes. you to the files page with that atlas selected.

<img width="821" alt="Screen Shot 2021-03-03 at 4 25 29 PM" src="https://user-images.githubusercontent.com/1334004/109874430-1d382480-7c3d-11eb-90fe-0091002b3805.png">

adamabeshouse:
where does the back button take you? @inodb

inodb:
@adamabeshouse back to explore, you get to this page when you are on the atlas page and click on an atlas name

PR #120, opened 2021-03-05
Atlas page adjustments

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/115

(1) Add "back" and "view files" buttons
(2) Remove all except first three tabs

### DIFF
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -5,7 +5,8 @@ import {toArabic} from "roman-numerals";
 
 import { WPAtlas } from '../types';
 import {ExploreOptionType, ExploreSelectedFilter} from "./types";
-
+import {ExploreURLQuery} from "../pages/explore";
+import {AtlasURLQuery} from "../pages/atlas/[id]";
 
 export function getRecords(obj: any): any {
     if (_.isObject(obj) || _.isArray(obj)) {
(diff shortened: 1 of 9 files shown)


lib/helpers.ts
```diff
@@ -209,23 +210,61 @@ export function parseRawAssayType(t:string) {
...
 }
-export function parseSelectedFilters(selectedFiltersURLQueryParam:string|undefined):ExploreSelectedFilter[] | null {
+export function parseSelectedFiltersFromUrl(selectedFiltersURLQueryParam:string|undefined):ExploreSelectedFilter[] | null {
     if (selectedFiltersURLQueryParam) {
         return JSON.parse(selectedFiltersURLQueryParam);
     }
     return null;
 }
 
+function addQueryStringToURL(url:string, queryParams: { [key:string]:string|undefined }) {
```
adamabeshouse, line 223: @alisman as the URLWrapper mastermind, what do you think about all this url-handling code? feels a bit ad-hoc to me but I don't know if it's time to do it better yet

pages/atlas/[id].tsx
```diff
@@ -53,6 +54,30 @@ const PostContent: React.FunctionComponent<{
...
+                        onClick={()=> router.back()}
+                    >
+                        {"< Back"}
+                    </Button>
+                )}
+                <Button
+                    variant={"link"}
+                    size={"sm"}
+                    style={{fontSize:12}}
+                    onClick={()=>router.push(getExplorePageURL([{
```
Comment (adamabeshouse) [inline pages/atlas/[id].tsx:72]: @alisman @inodb @onursumer thoughts on this way of creating the filter option on the fly? feels kind of hacky to me but I can't think of a better way to abstract it out so we reuse the logic here and on the explore page, because the explore page logic is buried within multiple components

lib/helpers.ts
alisman: seems fine for now.  but where is actual URL encoding take place?

Issue #133, opened 2021-03-09
JSON Update Related Fixes 


inodb:
- [x] Atlas tab id does not match up with files ID. E.g. on atlas tab `Transition to Metastatic State: Lung Cancer, Pancreatic Cancer and Brain Metastasis` is HTA8. But if you search `HTA8_` in Files tab it instead shows the atlas name as `Omic and Multidimensional Spatial Atlas of Metastatic Breast Cancers`, which is HTA_9` on the atlas tab. See: https://data.humantumoratlas.org/explore?tab=atlas @adamabeshouse  (https://github.com/ncihtan/htan-portal/pull/144)

- [x] Download tab popup shows name of file (it should display the snyapseid for downloading) @onursumer 

- [x] On biospecimen tab `Atlas ID` shows up lowercase @adamabeshouse 

- [x] Undefined shows up in the all filters search @adamabeshouse (https://github.com/ncihtan/htan-portal/pull/144)

    <img width="239" alt="Screen Shot 2021-03-09 at 1 53 06 PM" src="https://user-images.githubusercontent.com/1334004/110522390-d6d04300-80de-11eb-9108-a93d331ecbc3.png">
   

- [x] In the assay type dropdown i only see `ScRNA-seqLevel1`, but there should be level 2 and 3. So something goes wrong with data type there. It would also be good to remove the `Level` postfix and just list the assay type. One can use the File Type filter instead to get the level @alisman @adamabeshouse 

- [x] Are the many missing values in the data or an issue with our transformations?

    
    
    <img width="317" alt="Screen Shot 2021-03-09 at 1 58 05 PM" src="https://user-images.githubusercontent.com/1334004/110522949-7e4d7580-80df-11eb-8e3f-f0005511334d.png">

    <img width="317" alt="Screen Shot 2021-03-09 at 1 57 25 PM" src="https://user-images.githubusercontent.com/1334004/110522905-7097f000-80df-11eb-8396-56fea59f4cdb.png">

    EDIT: Seems like it's Vanderbilt (might be imaging level 1), we should reach out to the center @inodb @vthorsson 

- [x] search for pancreas NOS in all filters ->  select -> crash @onursumer (seems fixed by #137)

@alisman @adamabeshouse @onursumer let's split these up at 2PM

schultzn:
Here is some more feedback on names and displays.. looks great!!

- [ ]  - "Biospecimen" tab name should be "Biospecimens" to match the plural of the other tabs
[...]

PR #134, opened 2021-03-09
Display synapse id in the download popup

onursumer:
Related to #133

file_download_id

### DIFF
--- a/components/FileTable.tsx
+++ b/components/FileTable.tsx
@@ -12,7 +12,7 @@ import { faDownload, faSearch } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 interface IFileDownloadModalProps {
-    filenames: string[];
+    files: Entity[];
     onClose: () => void;
     isOpen: boolean;
 }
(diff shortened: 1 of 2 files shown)


components/FileTable.tsx
```diff
@@ -12,15 +12,17 @@ import { faDownload, faSearch } from '@fortawesome/free-solid-svg-icons';
...
     onClose: () => void;
     isOpen: boolean;
 }
 
 const FileDownloadModal: React.FunctionComponent<IFileDownloadModalProps> = (
     props
 ) => {
-    const script = props.filenames.map((f) => `synapse get ${f}`).join('\n');
+    const script = props.files
+        .map((f) => `synapse get ${f.HTANDataFileID}`)
```
onursumer, line 24: @inodb can you confirm that `HTANDataFileID` is actually the synapse id?

onursumer: Hmm, I don't think `HTANDataFileID` is the right id. Synapse ids should start with `syn`. Actually we have a field called `synapseId` which is currently missing in our `Entity` model: https://github.com/ncihtan/htan-portal/blob/b695603fb1f5622e74245da6db66732bbfbefca6/lib/helpers.ts#L71

But this synapse id is incorrect for most of the files. There are a lot of `undefined`s `null`s  and `2-Processed`s. Should we assume that this will be fixed at some point?

inodb: @milen-sage do you happen to know why the synapse ids might be missing? Is that on purpose? I can dig further if you don't

inodb: @milen-sage Not sure about exact number but perusing the JSON it's lot of the `bts:ImagingLevel2` from `HTA11`. Looks like this one doesn't have an `entityId` column:

https://www.synapse.org/#!Synapse:syn25127098

inodb (approved):
Although much of the data still has missing synapseIds - the code in this PR itself is good to go

PR #136, opened 2021-03-10
Truncate filename column

### LINKED ISSUE #129: Improve appearance of field name cell in explore page files list
We need to decide with @inodb how handle this, with truncation and toolip, wider column, smaller text, etc.
image

- [ ] make link to synapse so one can download a single file easily @onursumer 
- [ ] remove the path (just basename)
- [ ] would be good to be able to see extension
- [ ] truncate the name (maybe tooltip that allows you to mouse into it). Show first 10 and last 5 symbols with ellipses in between

onursumer:
Fix #129

Synapse link is still missing. Need to merge #134 which exposes the `synapseId` field.

filename_truncate

### DIFF
--- a/components/FileTable.tsx
+++ b/components/FileTable.tsx
@@ -3,10 +3,11 @@ import { observer } from 'mobx-react';
 import React, { SyntheticEvent } from 'react';
 import { Button, Modal, Form } from 'react-bootstrap';
 import DataTable from 'react-data-table-component';
+import Tooltip from 'rc-tooltip';
 import _ from 'lodash';
 import classNames from 'classnames';
 
-import { Atlas, Entity } from '../lib/helpers';
+import { Atlas, Entity, getFileBase, truncateFilename } from '../lib/helpers';
 import { getDefaultDataTableStyle } from '../lib/dataTableHelpers';
 import { faDownload, faSearch } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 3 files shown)


lib/helpers.ts
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import { NextRouter } from 'next/router';
 import fetch from 'node-fetch';
+import pathParse from 'path-parse';
```
inodb, line 4: feels slightly overkill to include a path parser dependency to do `lastIndexOf('/')`? No big deal tho, maybe it'll be useful in the future

inodb (approved):
Thanks!

lib/helpers.ts
onursumer: good point. actually, there is also builtin node path module, probably better to use them instead of this one.

onursumer: removed the `path-parse` dependency, using builtin node path package now.

components/FileTable.tsx
```diff
@@ -80,10 +81,24 @@ export default class FileTable extends React.Component<IFileTableProps> {
                 selector: 'filename',
                 wrap: true,
                 sortable: true,
+                cell: (file: Entity) => {
+                    return (
+                        <Tooltip overlay={getFileBase(file.filename)}>
+                            <a
+                                target="_blank"
+                                href={`https://www.synapse.org/#!Synapse:${
+                                    /* TODO file.synapseId */ file.filename
```
onursumer, line 90: we need to merge #134 to be able to refer to `file.synapseId`

PR #141, opened 2021-03-10
Make atlas ids uppercase

onursumer:
Related to #133

### DIFF
--- a/components/BiospecimenTable.tsx
+++ b/components/BiospecimenTable.tsx
@@ -21,6 +21,7 @@ export const BiospecimenTable: React.FunctionComponent<IBiospecimenTableProps> =
         {
             name: 'Atlas ID',
             selector: 'atlasid',
+            format: (sample: Entity) => sample.atlasid.toUpperCase(),
             wrap: true,
             sortable: true,
         },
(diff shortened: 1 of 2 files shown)


components/BiospecimenTable.tsx
```diff
@@ -21,6 +21,7 @@ export const BiospecimenTable: React.FunctionComponent<IBiospecimenTableProps> =
         {
             name: 'Atlas ID',
             selector: 'atlasid',
+            format: (sample: Entity) => sample.atlasid.toUpperCase(),
```
onursumer, line 24: Should we fix this when processing the data instead?

inodb: good point - i can prolly tweak it in the python script

adamabeshouse: +1 - it's been confusing for me to be dealing with casing in the frontend

alisman:
@onursumer @inodb shall we close this?

components/BiospecimenTable.tsx
inodb: fixed it here: https://github.com/ncihtan/htan-portal/pull/118

onursumer: Should we just close this PR without merging?

PR #144, opened 2021-03-12
Fix off-by-one error in data processing

adamabeshouse:
Fixes two points in https://github.com/ncihtan/htan-portal/issues/133:

"Atlas tab id does not match up with files ID"
and
"Undefined shows up in the all filters search"


inodb (approved):
LGTM - looking at the website at least

PR #150, opened 2021-03-16
Compress atlas id column

### LINKED ISSUE #146: Compress atlas id column on explore page
image

onursumer:
Fix #146


onursumer:
We may want to close this PR without merging. Seems like we are still not sure if we want to show name or id. See #159.

inodb:
@onursumer can we maybe add some column selection button? It would be easier to change the default once we have that and would be a useful feature anyway. Basically column selection and search for each tab

PR #167, opened 2021-03-18
Fix use effect infinite loop

### LINKED ISSUE #153: Scroll on main page is slow and choppy
Scrolling is slow and choppy when viewing the different atlases on the main page

https://data.humantumoratlas.org/

onursumer:
Fix #153

See https://stackoverflow.com/a/53074436


inodb (approved):
Nice - that solves it indeed! Thanks!

Issue #130, opened 2021-03-09
Support level 4 files


alisman:
Level 4 files are not currently shown because we don't have support for scenario when a file contains data from multiple participants and could therefor have multiple organs diagnosis etc.  Filters need to be adjusted to handle this.

adamabeshouse:
sorry what is level 4?

PR #169, opened 2021-03-22
Support level 4 files in file table

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/130


components/FileTable.tsx
```diff
@@ -103,15 +107,14 @@ export default class FileTable extends React.Component<IFileTableProps> {
             {
                 name: 'Biospecimen',
                 selector: (file: Entity) => {
-                    return file.primaryParents
-                        ? _.uniq(
-                              file.primaryParents.map(
-                                  (_file) =>
-                                      _file.biospecimen!.HTANBiospecimenID
-                              )
+                    const biospecimens = file.primaryParents
```
alisman, line 110: we shouldn't optimize as part of this PR but i think we should do this computation once up front for all files

adamabeshouse: good point

PR #184, opened 2021-03-24
Optimize explore page

adamabeshouse:
Time between loader appearing and UI appearing is down to ~1.8s from ~5.3s

### DIFF
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -100,16 +100,25 @@ export interface LoadDataResult {
 
 win.missing = [];
 
-function findPrimaryParents(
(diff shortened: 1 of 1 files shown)


lib/helpers.ts
```diff
@@ -100,15 +100,26 @@ export interface LoadDataResult {
 
 win.missing = [];
 
+function doesFileHaveMultipleParents(file: Entity) {
```
adamabeshouse, line 103: super weird .. I don't know why this happened in git. but anyway, its fine now

alisman (approved):

PR #188, opened 2021-03-26
Make FileTable search based on all components, and fix search input lagginess

adamabeshouse:
Next PR will be searching all tables


inodb (approved):
Awesome! Thanks!

Issue #191, opened 2021-03-30
Search Bug


inodb:
On atlas Tab

- [x] description is not included in search (maybe on purpose?)

On files Tab
- [x] Somehow couldn't search for HTA11 where 11 is any number

PR #195, opened 2021-03-30
EnhancedDataTable enhancements

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/191

(1) Add additionalSearchFilter to EnhancedDataTable, and use it to reimplement case id filtering in FileTable
(2) Add getSearchValue to EnhancedDataTable column specification
(3) Fix atlas table description selector


components/WPAtlasTable.tsx
```diff
@@ -31,7 +31,7 @@ export const WPAtlasTable: React.FunctionComponent<IWPAtlasTableProps> = (
         },
         {
             name: 'Atlas Description',
-            selector: 'title.rendered',
+            selector: 'WPAtlas.title.rendered',
```
onursumer, line 34: good catch! we should probably use function selectors for nested properties to avoid these kind of errors in the future.

adamabeshouse: good point @onursumer

PR #200, opened 2021-03-30
Add download instructions for level 1 and level 2 data

### LINKED ISSUE #160: For Level 1 and 2 Sequencing Data, users should be directed to CDS Instructions
Currently for Level 1 and 2 Sequencing data, you can click the Download button and you are directly to download these files from Synapse.  However, we cannot distribute these via Synapse, bc they are controlled access.  Instead, we should provide instructions on how to download from CDS.  Also flagging @milen-sage on this one, as he may know the CDS details.

onursumer:
Fix #160

### DIFF
--- a/components/FileTable.tsx
+++ b/components/FileTable.tsx
@@ -12,7 +12,12 @@ import {
 } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
-import { Entity, getFileBase, truncateFilename } from '../lib/helpers';
+import {
+    doesFileIncludeLevel1OrLevel2SequencingData,
+    Entity,
+    getFileBase,
+    truncateFilename,
+} from '../lib/helpers';
 import {
     getDefaultDataTableStyle,
     truncatedTableCell,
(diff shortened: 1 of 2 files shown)


components/FileTable.tsx
```diff
@@ -23,36 +27,84 @@ interface IFileDownloadModalProps {
...
+
+const FileDownloadModal: React.FunctionComponent<IFileDownloadModalProps> = (
+    props
+) => {
+    const cdsFiles = props.files.filter(
+        (f) => f.level === 'Level 1' || f.level === 'Level 2'
+    );
+    const synapseFiles = props.files.filter(
+        (f) => f.level !== 'Level 1' && f.level !== 'Level 2'
+    );
```
inodb, line 95: @onursumer in theory there could be imaging level 1 data that you prolly would download from synapse (it's only the sequencing data that one would download from CDS). So any Assay type that's not "Imaging"

onursumer: @inodb thanks for the clarification!

alisman (approved):

adamabeshouse (approved):

Issue #196, opened 2021-03-30
improve presentation of biospecimen list


alisman:
the wrapped specimen id's looks kinda crappy. lets discuss ways to improve

alisman:
I'm thinking we show "N specimen" and then do tooltip with vertical list of ids?

PR #220, opened 2021-04-06
Improve visualization of long lists of biospecimens in FileTable

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/196

image


alisman:
@adamabeshouse hah.  yeah, this is pretty ugly.  @inodb is there really a use case for someone actually needing to see those ids?

inodb:
@alisman Good question - maybe not really at this stage but at least for us it's useful to confirm the search by ID is working as expected :)

In the future we might want to be able to see what files were used to generate the level 4 or select the biospecimens/cases related to that one file

I think this is good enough for now though, just the border seems kinda thick

Issue #224, opened 2021-04-06
Change filter buttons at the top 


inodb:
- [x] Add a new filter called Atlas as the first one
- [x]  Rename Tissue Type → Organ, and move it to right of Atlas, before Cancer Type.
- [x] Change Cancer Type to Disease Type

Issue #223, opened 2021-04-06
Improve Search all filters pulldown


inodb:
Tweak the “Search all filters” pull down.

Currently searchable:

Atlas → keep as is
Organ → keep as is.
Assay → keep as is

- [x] Diagnosis → rename to match the other search box (cancer type)
- [x] Stage → remove for now
- [x] [Nice to have] Add in support for:
    - [x] Level
    - [x] File format

PR #227, opened 2021-04-06
Changes to filter controls

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/223
Fixes https://github.com/ncihtan/htan-portal/issues/224


inodb (approved):
Thanks 🙏 !! Looking at this some more:

- maybe good to have fixed width for these filters, might be a little easier on the eye
- If we want to save space we could remove "Type" from the filters to make them slightly shorter e.g. `Disease` -> `Disease Type`

This anyway is a great improvement on current production, so gonna merge

PR #234, opened 2021-04-08
Fix case where biospecimen has biospecimen as parent

### LINKED ISSUE #225: HTAPP cases are not showing up
Screen Shot 2021-04-06 at 2 13 52 PM

inodb:
Fix #225. HTAN Parent ID can be both participant or biospecimen. See also
https://data.humantumoratlas.org/standard/biospecimen

Hard to test right now since we removed HTAPP, but pre-removal this changed fixed it. Seems like it is the only atlas that has biospecimina as children of biospecimina

### DIFF
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -198,9 +198,14 @@ function getSampleAndPatientData(
                     biospecimenByHTANBiospecimenID[HTANParentID];
                 HTANParentID = parentBioSpecimen.HTANParentID;
             }
-            return diagnosisByHTANParticipantID[HTANParentID] as
-                | Entity
-                | undefined;
+            if (!(HTANParentID in diagnosisByHTANParticipantID)) {
+                console.error(
+                    `${s.HTANBiospecimenID} does not have a HTANParentID with diagnosis information`
+                );
+                return undefined;
+            } else {
+                return diagnosisByHTANParticipantID[HTANParentID] as Entity;
+            }
         })
         .filter((f) => !!f) as Entity[];
 
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -194,7 +194,7 @@ function getSampleAndPatientData(
             // going up the tree until participant is found.
             let HTANParentID = s.HTANParentID;
             while (HTANParentID in biospecimenByHTANBiospecimenID) {
-                let parentBioSpecimen =
+                const parentBioSpecimen =
                     biospecimenByHTANBiospecimenID[HTANParentID];
                 HTANParentID = parentBioSpecimen.HTANParentID;
             }
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -189,12 +189,19 @@ function getSampleAndPatientData(
         .filter((f) => !!f) as Entity[];
 
     const diagnosis = biospecimen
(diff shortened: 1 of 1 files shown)


lib/helpers.ts
```diff
@@ -189,12 +189,19 @@ function getSampleAndPatientData(
...
-                diagnosisByHTANParticipantID[s.HTANParentID] as
-                    | Entity
-                    | undefined
-        )
+        .map((s) => {
+            // HTANParentID can be both participant or biospecimen, so keep
+            // going up the tree until participant is found.
+            let HTANParentID = s.HTANParentID;
+            while (HTANParentID in biospecimenByHTANBiospecimenID) {
+                let parentBioSpecimen =
```
onursumer, line 197: I guess this one can be `const`
```suggestion
                const parentBioSpecimen =
```

lib/helpers.ts
```diff
@@ -189,12 +189,19 @@ function getSampleAndPatientData(
...
-            (s) =>
-                diagnosisByHTANParticipantID[s.HTANParentID] as
-                    | Entity
-                    | undefined
-        )
+        .map((s) => {
+            // HTANParentID can be both participant or biospecimen, so keep
+            // going up the tree until participant is found.
+            let HTANParentID = s.HTANParentID;
+            while (HTANParentID in biospecimenByHTANBiospecimenID) {
```
alisman, line 196: would this be more clear if straight up testing for participant identity instead of testing to see it is NOT a biospecimen?

adamabeshouse: I had the same thought. one counterpoint - this is a bit safer because it's testing membership in biospecimenByHTAN.. before accessing that same map

adamabeshouse (approved):

lib/helpers.ts
inodb: if the data is incorrect it could happen that `HTANParentID` is never in `diagnosisByHTANParticipantID`

will add a check for that

inodb: how about that (see below). Added some logging. Not sure if that's your recommended logging strategy. But it works, also logs known issues with HTAPP

adamabeshouse: I think logging is a good idea. I just noticed something else - the map is indexed by HTANBiospecimenID, but we're using HTANParentID to index into it. a bit confusing. is there another name that we can use that covers both?

inodb: yeah I was considering to `s/HTANParentID/biospecimenOrParticipantID`, but ofc further down the if/else logic you know which one it is, so the naming makes less sense. In some ways `HTANParentID` is actually kind of clear in that it could be both (tho it's maybe not obvious that it'll e.g. never by a FileID). The naming of the maps themselves make sense to me because they are indexed by only one type of ID

adamabeshouse: the only thing thats not clear is that an HTANParentID can also be an HTANParticipantID or an HTANBiospecimenID. like, its not clear that those are from the same set of ids. im not sure what the solution is, I agree renaming the map to be so verbose may actually be more confusing. I think the issue to resolve is how to clarify that these ids come from the same set *sometimes*

alisman (approved):

PR #235, opened 2021-04-08
Generate columns by data schema

### LINKED ISSUE #216: Autogenerate Biospecimen and Cases table columns
Use the data model to figure out what all the required and optional columns are. We could e.g. show all required columns by default.

onursumer:
Fix #216

### DIFF
--- a/components/BiospecimenTable.tsx
+++ b/components/BiospecimenTable.tsx
@@ -1,8 +1,12 @@
 import _ from 'lodash';
 import React from 'react';
+import { DataSchemaData, SchemaDataId } from '../lib/dataSchemaHelpers';
 import {
+    generateColumnsByDataSchema,
+    getAtlasColumn,
     getDefaultDataTableStyle,
     sortByBiospecimenId,
+    sortByHtanParentId,
 } from '../lib/dataTableHelpers';
 import { Atlas, Entity } from '../lib/helpers';
 import EnhancedDataTable from './EnhancedDataTable';
@@ -10,36 +14,37 @@ import EnhancedDataTable from './EnhancedDataTable';
 interface IBiospecimenTableProps {
     samples: Entity[];
     synapseAtlases: Atlas[];
+    schemaDataById?: { [schemaDataId: string]: DataSchemaData };
 }
 
 export const BiospecimenTable: React.FunctionComponent<IBiospecimenTableProps> = (
     props
 ) => {
-    const atlasMap = _.keyBy(props.synapseAtlases, (a) => a.htan_id);
-
-    const columns = [
+    const columns = generateColumnsByDataSchema(
+        SchemaDataId.Biospecimen,
+        props.schemaDataById,
+        // need to add a custom sort function for the id
         {
-            name: 'HTAN Biospecimen ID',
-            selector: 'HTANBiospecimenID',
-            wrap: true,
-            sortable: true,
-            sortFunction: sortByBiospecimenId,
-        },
-        {
-            name: 'Atlas Name',
-            selector: (sample: Entity) => {
-                return atlasMap[sample.atlasid].htan_name;
+            HTANBiospecimenID: {
+                sortFunction: sortByBiospecimenId,
             },
-            wrap: true,
-            sortable: true,
+            HTANParentID: {
+                sortFunction: sortByHtanParentId,
+            },
         },
-        {
-            name: 'Biospecimen Type',
-            selector: 'BiospecimenType',
-            wrap: true,
-            sortable: true,
-        },
-    ];
+        // Component seems to be always "Biospecimen", no need to have a column for it
+        ['Component']
+    );
+    const indexOfHtanBiospecimenId = _.findIndex(
+        columns,
+        (c) => c.id === 'HTANBiospecimenID'
+    );
+    // insert Atlas Name right after HTAN Biospecimen ID
+    columns.splice(
+        indexOfHtanBiospecimenId + 1,
+        0,
+        getAtlasColumn(props.synapseAtlases)
+    );
 
     return (
         <EnhancedDataTable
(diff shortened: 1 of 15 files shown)


lib/dataTableHelpers.tsx
```diff
@@ -115,3 +118,43 @@ export function sortByBiospecimenId(rowA: Entity, rowB: Entity) {
 
     return defaultNumericalComparison(rowA, rowB, iteratees);
 }
+
+export function generateColumnsByDataSchema<T>(
+    schemaId: string,
+    schemaDataMap?: { [id: string]: DataSchemaData },
```
adamabeshouse, line 124: can you name the `id` parameter more specifically here? this function is pretty abstract so careful naming would be helpful for readability

components/BiospecimenTable.tsx
```diff
@@ -33,12 +28,18 @@ export const BiospecimenTable: React.FunctionComponent<IBiospecimenTableProps> =
...
+            'bts:Biospecimen',
+            props.schemaDataMap,
+            // need to add a custom sort function for the id
+            {
+                HTANBiospecimenID: {
+                    sortFunction: sortByBiospecimenId,
+                },
+            },
+            // Component seems to be always "Biospecimen", no need to have a column for it
+            ['Component']
```
adamabeshouse, line 41: looks like the columns will be reordered, so now HTANBiospecimenId is no longer the first column, right? its okay?

components/ExploreTabs.tsx
```diff
@@ -8,13 +8,15 @@ import BiospecimenTable from './BiospecimenTable';
...
 import WPAtlasTable from './WPAtlasTable';
+import { DataSchemaData } from '../lib/dataSchemaHelpers';
 
 interface IExploreTabsProps {
     router: NextRouter;
     filteredFiles: Entity[];
     samples: Entity[];
     cases: Entity[];
     wpData: WPAtlas[];
+    schemaDataMap?: { [id: string]: DataSchemaData };
```
adamabeshouse, line 19: can the name be made any better? `schemaDataMap` is kind of confusing. for that matter, maybe it would be good to rename `DataSchemaData`, unless other people don't find that confusing? what is schema data?

adamabeshouse (changes requested):
code looks good but I think theres some names that it would be helpful to clarify

lib/dataTableHelpers.tsx
onursumer: We can name it `dataSchemaId`. We already use these ids in data standard pages. For example:

https://github.com/ncihtan/htan-portal/blob/0608bc381ccbd2723e3b3f550ac68bee74a5f039/pages/standard/biospecimen.tsx#L14-L16 

Actually, we can add an enum like `DataSchemaId` for the ones that we use in the code instead of using hardcoded strings.

components/BiospecimenTable.tsx
onursumer: Right, if the column order is important we will need to do some manual sorting.

components/ExploreTabs.tsx
onursumer: Data model and most of the utility functions were copied from the `Synapse-React-Client` repo. 

https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/01fb7e594517f9f4b4e9950fd0bd21e45e17eeb9/src/lib/containers/data_schema/types/IDataSchemaTypes.ts#L34

We should probably keep `DataSchemaData` for now however it sounds confusing. We can rename `schemaDataMap` to something better though. Maybe something like `schemaDataBySchemaId`?

components/CaseTable.tsx
```diff
@@ -31,20 +26,23 @@ export const CaseTable: React.FunctionComponent<ICaseTableProps> = (props) => {
...
-        {
-            name: 'Age at Diagnosis (years)',
-            selector: 'AgeatDiagnosis',
-            format: (sample: Entity) =>
-                convertAgeInDaysToYears(sample.AgeatDiagnosis),
-            wrap: true,
-            sortable: true,
-        },
+        ...generateColumnsByDataSchema(
+            SchemaDataId.Diagnosis,
```
onursumer, line 30: @inodb Do you know if `Diagnosis` is this the correct schema for Cases table?

lib/dataTableHelpers.tsx
onursumer: Added a new enum for clarity

onursumer:
The PR is ready to review. Both Case and Biospecimen tables are now auto generated from the schema. We may want to further customize the visible columns and the column order.

components/CaseTable.tsx
inodb: yes, though there could be multiple. But we can improve over time. E.g. there are a few others that might be at the "Cases" level

lib/dataSchemaHelpers.ts
```diff
@@ -72,6 +72,31 @@ export interface SchemaJson extends BaseEntity {
     '@graph': SchemaData[];
 }
 
+export enum SchemaDataId {
```
alisman, line 75: it seems weird that we're dynamically generating column but then hardcoding this?

inodb (approved):
From product side this looks good to me. Nice work! Lots of hidden data :)

Ordering of columns looks reasonable. But maybe good to keep `HTAN Participant ID` and `HTAN Biospecimen ID` as the first one?

lib/dataSchemaHelpers.ts
adamabeshouse: I disagree based on the usage - there is hardcoded stuff relating to specific data ids so its good to put it in an enum

inodb:
Maybe we can add description as the tooltip for the column? Can be follow up item

onursumer:
> Maybe we can add description as the tooltip for the column? Can be follow up item

Good idea! I was actually thinking the same thing, but somehow forgot about implementing :)

lib/dataSchemaHelpers.ts
onursumer: Right, for now we only enumerate the ids that we need to use, not all of them. We need to know the actual id when we want to customize a specific column.

PR #239, opened 2021-04-09
Add google analytics

inodb:
Add google analytics. I followed this example: https://github.com/vercel/next.js/tree/canary/examples/with-google-analytics. Account exists in cbioportal analytics account as discussed with @ecerami 

It is not disabled for local development. But maybe that's fine. In their example they set the analytics id as env variable on vercel: https://github.com/vercel/next.js/blob/canary/examples/with-google-analytics/lib/gtag.js#L1. For local development it's set as https://github.com/vercel/next.js/blob/canary/examples/with-google-analytics/.env.local.example

### DIFF
--- a/lib/gtag.js
+++ b/lib/gtag.js
@@ -1,0 +1,17 @@
+export const GA_TRACKING_ID = 'G-N6EV8C84KY';
+
+// https://developers.google.com/analytics/devguides/collection/gtagjs/pages
+export const pageview = (url) => {
+    window.gtag('config', GA_TRACKING_ID, {
+        page_path: url,
+    });
+};
+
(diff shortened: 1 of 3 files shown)


alisman:
@inodb coupla things.  
this fires a pageview event every time user changes filter on explore page (because we are updating route).  probably not great.  we might well want to track what people are filtering on but i don't think this will accomplish that well.  
also, currently we deal with tabs on explore page using qstring param.   probably would be better to use a page route so that they show up nicely as different page hits.

inodb:
@alisman Thanks for review! As discussed on slack: since we only need unique visitor counts for now this'll do. I'll file a ticket to sort the pageview thing

PR #254, opened 2021-04-14
Add tooltip for column headers

### LINKED ISSUE #245: Add tooltip to the column headers
Use the description from the schema for each column.

onursumer:
Fix #245

### DIFF
--- a/components/CaseTable.tsx
+++ b/components/CaseTable.tsx
@@ -26,7 +26,10 @@ export const CaseTable: React.FunctionComponent<ICaseTableProps> = (props) => {
                 sortFunction: sortByHtanParticipantId,
             },
             AgeatDiagnosis: {
+                // we need to customize both the name and the tooltip since we convert days to years
                 name: 'Age at Diagnosis (years)',
+                headerTooltip:
+                    'Age at the time of diagnosis expressed in number of years since birth.',
                 format: (sample: Entity) =>
                     convertAgeInDaysToYears(sample.AgeatDiagnosis),
             },
@@ -36,7 +39,7 @@ export const CaseTable: React.FunctionComponent<ICaseTableProps> = (props) => {
     );
     const indexOfHtanParticipantId = _.findIndex(
         columns,
-        (c) => c.id === 'HTANParticipantID'
+        (c) => c.id === 'HTAN Participant ID'
     );
     // insert Atlas Name right after HTAN Participant ID
     columns.splice(
--- a/components/EnhancedDataTable.tsx
+++ b/components/EnhancedDataTable.tsx
@@ -10,6 +10,7 @@ import DataTable, {
 import {
     resolveColumnVisibility,
     getColumnVisibilityMap,
+    getColumnKey,
 } from '../lib/dataTableHelpers';
 import DebouncedObservable from '../lib/DebouncedObservable';
 import { ColumnVisibility } from './ColumnSelect';
@@ -90,14 +91,15 @@ function getColumnVisibility<T>(
 
     (columns || [])
         .filter((column) => column.name)
-        .forEach((column) =>
+        .forEach((column) => {
+            const columnKey = getColumnKey(column);
             colVisProp.push({
-                id: column.name!.toString(),
-                name: column.name!.toString(),
-                visible: columnVisibility[column.name!.toString()],
+                id: columnKey,
+                name: columnKey,
+                visible: columnVisibility[columnKey],
                 toggleable: isColumnToggleable(column),
-            })
-        );
+            });
+        });
 
     return colVisProp;
 }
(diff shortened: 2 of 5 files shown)


components/EnhancedDataTable.tsx
```diff
@@ -182,7 +184,7 @@ export default class EnhancedDataTable<T = any> extends React.Component<
     get columns(): IDataTableColumn[] {
         return (this.props.columns || []).map((c) => ({
             ...c,
-            omit: c.name ? !this.columnVisibility[c.name.toString()] : c.omit,
+            omit: c.name ? !this.columnVisibility[getColumnKey(c)] : c.omit,
```
onursumer, line 187: In case `name` is a component `toString` does not really work. We should use the column key instead.

lib/dataTableHelpers.tsx
```diff
@@ -160,16 +164,22 @@ export function generateColumnsForDataSchema<T>(
...
-                        name: schema.attribute,
-                        selector: columnKey,
+                        id: schema.attribute,
+                        name: (
+                            <Tooltip
+                                overlay={<span>{schema.description}</span>}
+                            >
+                                <span>{schema.attribute}</span>
+                            </Tooltip>
+                        ),
```
onursumer, line 177: Making the name a component breaks the current column visibility implementation. So, this PR also have changes to address that problem.

inodb (approved):
Thanks! Looks good to me - one minor thing is that some of the first columns (participant id, Age) don't have a tooltip, so it's not immediately obvious that others do

components/EnhancedDataTable.tsx
```diff
@@ -92,9 +93,9 @@ function getColumnVisibility<T>(
         .filter((column) => column.name)
         .forEach((column) =>
             colVisProp.push({
-                id: column.name!.toString(),
-                name: column.name!.toString(),
-                visible: columnVisibility[column.name!.toString()],
+                id: getColumnKey(column),
```
adamabeshouse, line 96: I know its a super cheap function, but it makes me cringe a bit to see the function called 3 times

lib/dataTableHelpers.tsx
adamabeshouse: is it possible to rename the `name` attribute to `header` to reflect this?

adamabeshouse (changes requested):
small changes - see comments

lib/dataTableHelpers.tsx
onursumer: `name` is the property the `react-data-table-component` library is using for the header component. Changing this might be tricky at this point. We would still need to convert `header` to `name` under the hood. Not sure if it is worth it.

components/EnhancedDataTable.tsx
onursumer: Good catch! Fixed.

onursumer:
> Thanks! Looks good to me - one minor thing is that some of the first columns (participant id, Age) don't have a tooltip, so it's not immediately obvious that others do

@inodb Good point. Updated the PR to keep the tooltip when the column name is customized. `Atlas Name` still needs a custom tooltip though, since it is manually added outside the schema.

lib/dataTableHelpers.tsx
```diff
@@ -132,7 +136,12 @@ export function getAtlasColumn(atlases: Atlas[]) {
     const atlasMap = _.keyBy(atlases, (a) => a.htan_id);
 
     return {
-        name: 'Atlas Name',
+        id: 'Atlas Name',
+        name: (
+            <Tooltip overlay="Name of the Atlas">
```
onursumer, line 141: @inodb Can you think of a better description for Atlas Name?

lib/dataTableHelpers.tsx
adamabeshouse: okay, no problem

lib/dataTableHelpers.tsx
inodb: lol seems fine

components/CaseTable.tsx
```diff
@@ -36,7 +39,7 @@ export const CaseTable: React.FunctionComponent<ICaseTableProps> = (props) => {
     );
     const indexOfHtanParticipantId = _.findIndex(
         columns,
-        (c) => c.id === 'HTANParticipantID'
+        (c) => c.id === 'HTAN Participant ID'
```
alisman, line 42: we have a string with spaces as ID?

onursumer: Right. We cannot use the name property to get the actual name of the column anymore, because name is now a component not a string. Ideally we should have another property like header to format the column header, but the actual table library uses name to render the header. So this is a workaround to use the id as name. Not a perfect solution though.

PR #257, opened 2021-04-16
Make some optimizations to explore page

adamabeshouse:
1 - using webworker for one part of text filtering
2 - memoizing a commonly used function


onursumer (commented):
Interacting with filters usually causes the page to crash with an error message `Cannot read property 'htan_name' of undefined`

adamabeshouse:
I think aaron found that making `filteredFiles` `@computed` completely optimized it, so this is unnecessary (and apparently buggy)

PR #263, opened 2021-04-16
Fix duplicate bulk WES enum values

### LINKED ISSUE #259: Bulk WES Level 1 appears three times on the standards page
Screen Shot 2021-04-16 at 6 46 48 AM

onursumer:
Fix #259


inodb:
Thanks!!

PR #267, opened 2021-04-19
Add Demographics to the Cases table

### LINKED ISSUE #253: Can we add demographics to case tab as well?
Would be nice to be able to see `Demographics` info as well (e.g. Gender). Can we add the columns for this one to the cases tab?

See: https://data.humantumoratlas.org/standard/clinical

onursumer:
Fix #253

### DIFF
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -84,6 +84,8 @@ export interface Entity {
     WPAtlas: WPAtlas;
     biospecimen: Entity[];
     diagnosis: Entity[];
+    demographics: Entity[];
+    cases: Entity[];
     primaryParents?: Entity[];
     synapseId?: string;
 }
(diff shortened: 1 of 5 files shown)


lib/helpers.ts
```diff
@@ -84,6 +84,8 @@ export interface Entity {
     WPAtlas: WPAtlas;
     biospecimen: Entity[];
     diagnosis: Entity[];
+    demographics: Entity[];
+    cases: Entity[];
```
onursumer, line 88: There is some duplication here. `cases` is a field where we merge diagnosis and demographics together.

adamabeshouse: one way to handle this would be in typing, e.g. you could type it as `cases:Pick<Entity, whatevermembersarerelevant>[]`, etc. that would make it clearer what each is used for even if there's overlap

adamabeshouse (approved):

inodb (approved):
Sweet - thanks!

lib/helpers.ts
onursumer: Oh I meant the data itself is now duplicated. cases is basically "diagnosis + demographics". Maybe okay for now. It is shallow copy though, so maybe okay to duplicate.

adamabeshouse: how about a utility method `getCases(Entity)` -> `demographics.concat(diagnosis)` instead of the new member?

onursumer: We cannot simply concat them, we have to merge actual objects by htan participant id. The utility function for that one is here: https://github.com/ncihtan/htan-portal/blob/master/lib/helpers.ts#L238

PR #268, opened 2021-04-19
Remove Synapse links for level 1 and 2 data

### LINKED ISSUE #261: Remove Synapse links for level 1 and 2 data
Clicking on an individual level 1 or 2 file brings you to the Synapse page for that file. Users will access these data through CDS in the future.

onursumer:
Fix #261


inodb (approved):
Thanks!

PR #275, opened 2021-04-27
Increase filename col size

### LINKED ISSUE #272: wrapping of filename issue
Screen Shot 2021-04-23 at 4 41 50 PM

The filename wraps around, since we're truncating it, it would be nice if it would fit in a single line

onursumer:
Fix #272


inodb (approved):
Thanks!

PR #276, opened 2021-04-29
Adjust numerical columns to the right

### LINKED ISSUE #273: right justify columns with numbers
E.g. on the atlas page

Screen Shot 2021-04-23 at 4 51 30 PM

onursumer:
Fix #273

### DIFF
--- a/lib/dataSchemaHelpers.ts
+++ b/lib/dataSchemaHelpers.ts
@@ -105,7 +105,66 @@ export enum SchemaDataId {
     PancreaticCancerTier3 = 'bts:PancreaticCancerTier3',
     ProstateCancerTier3 = 'bts:ProstateCancerTier3',
     SarcomaTier3 = 'bts:SarcomaTier3',
+
+    AgeAtDiagnosis = 'bts:AgeatDiagnosis',
+    YearOfDiagnosis = 'bts:YearofDiagnosis',
+    DaysToLastFollowup = 'bts:DaystoLastFollowup',
+    DaysToLastKnownDiseaseStatus = 'bts:DaystoLastKnownDiseaseStatus',
+    DaysToDiagnosis = 'bts:DaystoDiagnosis',
+    PercentTumorInvasion = 'bts:PercentTumorInvasion',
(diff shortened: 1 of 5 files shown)


lib/dataSchemaHelpers.ts
```diff
@@ -105,8 +105,67 @@ export enum SchemaDataId {
...
+    [SchemaDataId.PercentInflamInfiltration]: true,
+    [SchemaDataId.PercentLymphocyteInfiltration]: true,
+    [SchemaDataId.PercentMonocyteInfiltration]: true,
+    [SchemaDataId.PercentNecrosis]: true,
+    [SchemaDataId.PercentNeutrophilInfiltration]: true,
+    [SchemaDataId.PercentNormalCells]: true,
+    [SchemaDataId.PercentStromalCells]: true,
+    [SchemaDataId.PercentTumorCells]: true,
+    [SchemaDataId.PercentTumorNuclei]: true,
+};
```
onursumer, line 167: We wouldn't need any of these if `DataSchemaData` has better values for the `type` field. Unfortunately it is almost always `rdfs:Class` which is not very helpful.

inodb: hmm yeah I think it would be good to fix the schema. @milen-sage is this something that would be easy to add?

inodb:
Thanks for fixing this! Noticed the alignment looks slightly funky on the atlas tab of the explore page:

Screen Shot 2021-04-29 at 8 18 03 AM

Might be resolved by right adjusting the headers too? Maybe viewers can be slightly more wide and Cases and Biospecimen more slim?

onursumer:
> Might be resolved by right adjusting the headers too? Maybe "Viewers" column can be slightly more wide and Cases and Biospecimen more slim?

@inodb Aligning only some of the column headers seems tricky (see https://github.com/jbetancur/react-data-table-component/issues/788). I will try to find a workaround for this.

onursumer:
@inodb found a workaround to align column headers only for Atlas table. Let me know if this is good to go.

inodb:
Thanks!!

PR #292, opened 2021-05-11
Fix #291 Only show images part of data release

### LINKED ISSUE #291: Only list 333 images part of first data release
Currently we are listing all images, only list the QC'ed images for first data release instead

Screen Shot 2021-05-11 at 11 49 57 AM

inodb:
There is an issue with vanderbilt's mxif manifest file (missing entityId
column). Once that is fixed, these code changes should result in the 333
image files.


dgutman:
Yes-- we will likely be updating some of the vanderbilt metadata.. I noticed some of the image sizes didn't appear consistent either.  So it should bounce back to 333 once we fix that image.  Also next update should have the ability to show more colorful thumbnails for the fluorescent images.

PR #296, opened 2021-06-01
Exclude specific HTAPP folders

inodb:
We have a list of HTAPP folders to include for release 1. Only include
those files. Add `click` command line interface to easily
include/exclude these folders from the synapse JSON.

### DIFF
--- a/data/get_syn_data.py
+++ b/data/get_syn_data.py
@@ -1,3 +1,4 @@
+import click
 import pandas as pd
 import numpy as np
 import wget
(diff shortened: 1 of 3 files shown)


inodb:
Going to merge this b/c some pending changes depend on this. The only files that are removed are IMC from WUSTL, but that is expected, since they are not in `image-release-1-synapse-ids.json`

Screen Shot 2021-06-01 at 4 22 54 PM

Feel free to open issues for anything else

Issue #302, opened 2021-06-15
Attempt to pre-compute the transformation of sage json dump


alisman:
In order to make file data serializable (and thus pre-computable at build time) we can no longer attach entities (e.g. diagnosis) to files as references.  We need instead to look them up at runtime.  

This may have a serious performance consequence because lookup in a table has to be slower than just accessing a reference.   Lets see. 

This precomputation can be done as part of the getStaticData hook of next.js (in explore.tsx page class) which executes at build time.

PR #303, opened 2021-06-16
Precompute SAGE dump transformation in getStaticProps

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/302

This required adjusting some of our transformations so that the data doesn't
take forever to serialize, and/or crash the server.

(1) We no longer duplicate the entire atlas object on each file, so
references require a lookup based on atlasid

(2) We no longer duplicate entities in the primaryParents member, so
references require a lookup based on HTANDataFileId

(3) Fix an issue where biospecimen, diagnosis, case, and demographic fields
were having tons of redundancy for entities with a lot of primaryParents


inodb (commented):
@adamabeshouse Thanks! Is this ready to review? I'm seeing some 500 error here https://htan-portal-nextjs-4l5i48eg9-htan.vercel.app/explore

PR #307, opened 2021-07-08
add demographics filter

inodb:
EDI working group asked for easy access to demographic related numbers, so added another filter

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 5000",
     "build": "next build",
     "start": "next start",
     "processSynapseJSON": "npx ncc run data/processSynapseJSON.ts",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev -p 5000",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 5000",
```
inodb, line 6: will need to remove before merging

inodb:
Identified some issue here, because we do filtering on files only, the cases tab doesn't always reflect the filters (e.g. if any of the cases is in a combined level 4 file). See e.g. MSK atlas which contains one african american case, but shows 50 in the stats as well as on the cases tab: 

Screen Shot 2021-07-13 at 10 23 51 AM


https://htan-portal-nextjs-cfgs1fry2-htan.vercel.app/explore?tab=cases&selectedFilters=%5B%7B%22value%22%3A%22black+or+african+american%22%2C%22label%22%3A%22black+or+african+american%22%2C%22group%22%3A%22Race%22%2C%22count%22%3A739%2C%22isSelected%22%3Afalse%7D%2C%7B%22value%22%3A%22HTAN+MSK%22%2C%22label%22%3A%22HTAN+MSK%22%2C%22group%22%3A%22AtlasName%22%2C%22count%22%3A279%2C%22isSelected%22%3Afalse%7D%5D

PR #323, opened 2021-09-29
Add charts to homepage

onursumer:
Fix https://github.com/ncihtan/htan-portal/issues/348


inodb:
@onursumer do you think we can use the same mapping that is used here? https://portal.gdc.cancer.gov/. E.g. when you click on "Colorectal" it gives a specific set of values for the filter. We are using the same data dictionary as them so might be nice if we can leverage their mapping. It might be here somewhere: https://github.com/NCI-GDC/portal-ui

onursumer:
@inodb Should we just have a simple `cases by organ` bar chart (or stacked bar chart `by atlas`) on the home page? We can then apply the appropriate filter upon clicking on a certain bar.

inodb:
Superseded by https://github.com/ncihtan/htan-portal/pull/355

Issue #319, opened 2021-09-14
Update numbers on atlases tab depending on filters


inodb:
- [ ] Currently numbers always reflect total cases + biospecimens. Let's adjust whenever a filter is selected. E.g. if you select Demographics > Female. It can show for each atlas in the cases column `x / total number of cases`

This is minor unrelated:

- [ ] The sorting on atlases tab is not clear. Maybe we should add the ID column back in, so it's clear that's what the sorting indicates?

PR #325, opened 2021-09-30
Show filtered case and sample counts in atlas table

adamabeshouse:
Fixes https://github.com/ncihtan/htan-portal/issues/319

image


inodb (approved):
Awesome 🥇 !!!

PR #326, opened 2021-10-04
Add e2e framework and initial spec

alisman:


alisman:
@adamabeshouse can you try installing and running this?
@inodb once we have this going, seems like it's relatively simple to use github actions to run these tests against the Vercel preview deployments.

PR #335, opened 2021-11-16
Upgrade next@12, react@17, node-sass@6, remove babel

inodb:
Was getting some out of memory errors: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/271. I then realized that nextjs actually recommends to use IDE's type checking instead: https://github.com/vercel/next.js/pull/15103/files. I'm sure there was a specific reason to use fork-ts-checker but not 100% sure if it is still necessary? I reverted the commit that introduced it for now

Second, this upgrades to next@12 which required some upgrades to react and node-sass. They also [no longer use babel](https://nextjs.org/docs/upgrading#swc-replacing-babel) so removed that as well (it was introduced when adding mobx)

TODO:
- [x] Do more thorough testing if all functionality is still there


adamabeshouse (approved):

onursumer:
@inodb IDE's type checking is not always consistent with the actual build errors, but I guess we can live with that.

inodb:
@onursumer I see - that makes sense. Yeah i couldn't get it to work without out of memory errors on my machine so maybe we can re-introduce it at some point if it turns out we're missing many of these typescript errors during development

inodb:
@onursumer @alisman I filed an issue to re-enable type checking in dev mode in a separate PR: https://github.com/ncihtan/htan-portal/issues/338

onursumer (approved):

PR #344, opened 2022-01-24
Integrate two HTAN sites

alisman:
Fix https://github.com/ncihtan/htan-portal/issues/347


inodb:
@alisman Starting to look nice! Wondering if we should maybe play around with the dropdowns a bit. Might be nice to do something like GDC: https://gdc.cancer.gov/about-gdc
Screen Shot 2022-01-25 at 1 15 49 PM

Note that their data portal is separate, I just mean that the dropdowns themselves look pretty neat and allow for a lot of subheaders

inodb:
superseded by https://github.com/ncihtan/htan-portal/pull/355

PR #345, opened 2022-01-25
Update asset mapping

adamjtaylor:
- New asset mapping file `data/htan-assets-latest.json` with flat structure and Synapse ID.
- `FileTable` now matches these based on Synapse ID over Filename which should be more robust


adamjtaylor:
Suggestions from @inodb 
>yeah looks good to me! Maybe these minor tweaks:
>- `data/htan-assets-latest.json` -> `data/htan-imaging-assets.json`
>- Can we JSON format that file? Can e.g. do `cat data/htan-assets-latest.json |python3 -m json.tool > data/htan-imaging-assets.json` . When there are data updates it’ll be easier to inspect the diff and e.g. see what new thumbnails/stories were added
>- remove `data/imaging-thumbnail-mappings.json`

inodb (approved):
Great - thanks so much!

PR #366, opened 2022-03-01
Fix matching of sample with parent caseId

alisman:


lib/helpers.ts
```diff
@@ -137,6 +137,17 @@ export function fillInEntities(data: LoadDataResult): Entity[] {
...
+    // give each biospecimen it's caseid (i.e "diagnosis" HTANParticipantID)
+    // biospecimen have HTANParentID but that may or may not be it's caseid because
+    // biospecimen can have other biospecimen as parents (one case at top)
+    _.forEach(data.biospecimenByHTANBiospecimenID, (specimen) => {
+        const parentIdMatch = specimen.HTANParentID.match(/[^_]*_[^_]*/);
+        // we should always have a match
+        specimen.HTANParticipantID =
+            specimen.HTANParticipantID ||
+            (parentIdMatch ? parentIdMatch[0] : '');
+    });
```
inodb, line 149: it seems like similar logic lives here:

https://github.com/ncihtan/htan-portal/blob/f2c54b71dda65cb620b4274589659434a4842a78/data/processSynapseJSON.ts#L272-L274

Not sure if we can unify this somehow

inodb (approved):
Thanks!!

lib/helpers.ts
inodb: filed an issue for this https://github.com/ncihtan/htan-portal/issues/367

PR #368, opened 2022-03-10
Add partial stanford release 2 data

inodb:
- remove duplicate entries from diagnosis/demographics etc on data pull
- also adds some new fields from schema (biospecimen data source related
to SRRS)
- ~scRNASeq level 2 excluded for now (some issues with parent data file ids https://sagebionetworks.jira.com/browse/HTAN-39)~


adamjtaylor:
I just spoke with Stanford. They are hoping to submit the corrected scRNA-seq data tonight so let's see if that goes OK then we can update this PR to include it if it looks OK.

adamjtaylor:
Re the duplicate @milen-sage and I need to take a closer look at both the current schema and the use case (same participant, multiple resections). For now we can continue to use the de-duplication as the rectosigmoid junction could be also be described as `Colon NOS`.

PR #418, opened 2022-07-07
Include TNP Sardana in get_syn_data script

adamjtaylor:
Per email from HMS, SARDAN's data is not listed on the portal. This PR adds them (back) into the get_syn_data script


inodb:
fixed btw, thanks for suggestion!

PR #422, opened 2022-07-27
Add BigQuery in tools tab of CHOP and Vanderbilt

leexgh:
For example:
https://htan-portal-nextjs-git-fork-leexgh-add-bigquerry-htan.vercel.app/publications/chop_all_chen_2022?tab=tools
image

https://htan-portal-nextjs-git-fork-leexgh-add-bigquerry-htan.vercel.app/publications/vanderbilt_crc_chen_2021?tab=tools
image


leexgh:
@inodb Could you help to confirm if the link of `cellxgene` logo in CHOP is correct?
image
https://humantumoratlas.org/explore?selectedFilters=%5B%7B%22group%22%3A%22AtlasName%22%2C%22value%22%3A%22HTAN+CHOP%22%7D%5D
It seems only CHOP cellxgene goes to portal explore page but other all link out to cellxgene website

inodb:
Thanks @leexgh !

- The CHOP cellxgene link is correct. We still need to get the proper combined dataset up on the cellxgene website (like for MSK), so for now it's just pointing to the files tab instead
-  Could you tweak the bigquery/minerva images a bit? Might be better to show an example of what bigquery looks like:
    Screen Shot 2022-07-29 at 4 50 10 PM
    Ok to just keep linking to the same page, but this way one can immediately get an approx idea about the tool rather than having to click twice

inodb (approved):
Thank you!!

PR #431, opened 2022-08-13
Add level filter on publications page files tab

leexgh:
Add level filter dropdown on publication page
For example:
https://htan-portal-nextjs-git-fork-leexgh-level-filter-htan.vercel.app/publications/chop_all_chen_2022?tab=scrna-seq
image


inodb (approved):
Great! Thanks so much!

PR #440, opened 2022-09-16
Update HTAN Data Standards Docs and MetaData Tables

ecerami:
Thorough updating of the data standards pages to:

1. Match the correct version of the HTAN JSON LD model.
2. Match the new HTAN Missing Manual

(this is also my first PR in a very long time.  so let me know if I messed anything up!)

Additional changes (inodb edit):

- Removes labels column. This is only used to get the field attribute name, but maybe not immediately necessary for a reader of data standards page.
    Screen Shot 2022-09-16 at 3 41 35 PM
- Removes labeling of what columns are required/optional. This could be relevant for a data submitter but there might be a cleaner way to show this
- Disables "show more" button. Prolly not super necessary to be able to see all values (subset give you enough of a clue)
    This PR:
    Screen Shot 2022-09-16 at 3 45 48 PM
   used to be:
    Screen Shot 2022-09-16 at 3 44 15 PM
 - The schema update removes a column from biospecimens tab (`Site of resection or Biopsy`) and introduces a new hidden one (`Source HTAN Biospecimen ID`)


inodb (approved):
Thanks so much @ecerami ! Looks great to me! You still got it 🏆!

I made some edits to description of the PR to mention some other changes I noticed in the PR. Feel free to edit

components/DataSchema.tsx
```diff
@@ -78,34 +79,35 @@ const DataSchemaTable: React.FunctionComponent<{
...
-                const text = getDataSchemaValidValues(
+                let text = getDataSchemaValidValues(
                     schemaData,
                     props.dataSchemaMap
                 )
-                    .map((s) => s.label)
-                    .join(', ');
-
+                    .map((s) => s.attribute)
+                    .join('#');
```
onursumer, line 106: it's not immediately clear here why we join with `#`. Maybe add a comment to indicate that we use `#` as a special delimiter later for formatting purposes.

lib/dataTableHelpers.tsx
```diff
@@ -22,6 +22,14 @@ export function getDefaultDataTableStyle() {
...
             style: {
                 fontSize: 14,
+                lineHeight: '1.5em',
+                display: 'block',
+            },
+        },
+        rows: {
+            style: {
+                paddingTop: '10px',
+                paddingBottom: '10px',
```
onursumer, line 32: Just a heads up. This will affect every data table we have for the entire web site. Not just the data standards page.

components/ExpandableText.tsx
```diff
@@ -34,26 +34,39 @@ const ExpandableText: React.FunctionComponent<ExpandableTextProps> = (
...
+    for (let token of tokens) {
+        if (token.length > 0) {
+            options.push(
+                <span>
+                    - {token.toLowerCase()}
+                    <br />
+                </span>
+            );
+        }
+    }
```
onursumer, line 49: I would use filter and map combination here. Something like

```suggestion
    const options = tokens.filter(token => token.length > 0).map(token => 
        <span>
            - {token.toLowerCase()}
            <br />
        </span>
    );
```

components/DataSchema.tsx
```diff
@@ -78,34 +79,35 @@ const DataSchemaTable: React.FunctionComponent<{
         },
     ];
 
+    // Remove Required Column for now.f
```
onursumer, line 82: why the f? :)
```suggestion
    // Remove Required Column for now.
```

onursumer (changes requested):
Added a few minor change request. Other than that it looks pretty good to me. Good job 🙌

onursumer:
Looks like this PR changes the formatting for `Organ` and `Diagnosis` columns of the file table. Probably not intended.

Here is the change:

files_organ diagnosis_pr

This is the current version on production:

files_organ diagnosis_production

onursumer:
Also looks like the checkboxes of the file table are not properly aligned anymore:

files_checkbox_misaligned

https://htan-portal-nextjs-6p6u6vpwt-htan.vercel.app/explore?tab=file

ecerami:
Thanks @onursumer!  great suggestions.

dumb question for you:  I have a branch with all these changes.  if I make the code changes you suggest, and commit them to the branch, do those changes automatically become part of the PR?

components/DataSchema.tsx
ecerami: Good catch.  Now fixed:  https://github.com/ncihtan/htan-portal/pull/440/commits/7f9469481f29ab7b4b6f465ab9946a94298eeda6

styles/app.scss
```diff
@@ -178,6 +178,19 @@ table {
     }
 }
 
+.table_50 {
```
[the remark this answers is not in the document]
ecerami: Resolved here:  https://github.com/ncihtan/htan-portal/pull/440/commits/986e84226b51c3a581bced1a3156f4fb1b018905

components/DataSchema.tsx
```diff
@@ -78,34 +79,35 @@ const DataSchemaTable: React.FunctionComponent<{
...
+    // }
 
     // conditionally show valid values column
     if (hasNonEmptyValidValues(props.schemaData)) {
         columns.push({
             name: 'Valid Values',
             selector: 'validValues',
             cell: (schemaData: DataSchemaData) => {
-                const text = getDataSchemaValidValues(
+                let text = getDataSchemaValidValues(
```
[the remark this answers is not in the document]
ecerami: Changed let back to const.  See:  https://github.com/ncihtan/htan-portal/pull/440/commits/496712fc830cb58a92245f74a20b54e5eb2efc89

components/DataSchema.tsx
ecerami: Added comments, as you suggested.  See https://github.com/ncihtan/htan-portal/pull/440/commits/496712fc830cb58a92245f74a20b54e5eb2efc89

components/ExpandableText.tsx
ecerami: fancy :-)

ecerami: @onursumer that's some nice code :-)

Added here:  https://github.com/ncihtan/htan-portal/pull/440/commits/280b2c4723f0b80610574f04a4152bd9c3ee9bb8

components/ExpandableText.tsx
```diff
@@ -34,26 +34,39 @@ const ExpandableText: React.FunctionComponent<ExpandableTextProps> = (
         setTruncated(!truncated);
     };
 
+    // Disable toggler for now
+    let options: Array<JSX.Element> = [];
+    let tokens = props.fullText.split('#');
```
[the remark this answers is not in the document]
ecerami: Fixed here:  https://github.com/ncihtan/htan-portal/pull/440/commits/280b2c4723f0b80610574f04a4152bd9c3ee9bb8

lib/dataTableHelpers.tsx
ecerami: ah, did not realize that.

ecerami:
Hi @onursumer:  I think I managed to address most of you issues.  Except one...  It seems that we use the ExpandableText in the File Table.  Do we need the more/less functionality there?  I had assumed that ExpandableText was only used in the standards page.  But, if we use it multiple other places, then maybe I need to roll back my ExpandableText changes?

onursumer:
> Hi @onursumer: I think I managed to address most of your issues. Except one... It seems that we use the ExpandableText in the File Table. Do we need the more/less functionality there? I had assumed that ExpandableText was only used in the standards page. But, if we use it multiple other places, then maybe I need to roll back my ExpandableText changes?

@ecerami I think what we should do is to keep the functionality of `ExpandableText` as is, and add a new component for DataStandards page. With the changes introduced in this PR `ExpandableText` becomes not expandable.

ecerami:
@onursumer I agree, but unfortunately, I don't know the code base well enough to do that.  is this something you can help do?

onursumer:
> @onursumer I agree, but unfortunately, I don't know the code base well enough to do that. is this something you can help do?

@ecerami sure, I can help with this.

ecerami:
thanks @onursumer!  much better that you make those changes :-)

ecerami:
@inodb @onursumer:  ok, I think we are good now.  I addressed all of @onursumer's suggestion, except the more complicated one, which he now did :-). thanks @onursumer!!  and, thanks @inodb for all your help on this one.

inodb:
Thanks again for the PR! We saw one more issue with the data table styles getting applied everywhere during reviewing, which @onursumer will follow up on

inodb:
Ok looks good now! Made one minor change to the description and @onursumer fixed everything else 🥇 !

ecerami:
Hooray! thanks @inodb and @onursumer

PR #444, opened 2022-10-14
add download source filter

inodb:
Screen Shot 2022-10-14 at 6 09 33 AM

Testing URL:

TODO:

- [ ] I think my dbGaP synapse id list is not up to date, there are a few that I see on seven bridges
- [ ] Many image files are missing from IDC (is this really correct, or since they are now transferred differently I am missing some?). I'm using a list that IDC shared before
- [ ] Need to investigate the non downloadable files further


lib/helpers.ts
```diff
@@ -60,6 +60,8 @@ export interface BaseSerializableEntity {
     WPAtlas: WPAtlas;
     primaryParents?: HTANDataFileID[];
     synapseId?: string;
+    isRawSequencing?: boolean;
+    downloadSource?: 'dbGaP' | 'IDC' | 'Synapse' | 'Coming Soon';
```
onursumer, line 64: might be worth adding a `DownloadSource` enum. I see these values hard-coded in multiple places.

inodb: Good point! Added an enum - lmk if that looks ok

onursumer (approved):
🙌

PR #449, opened 2022-11-01
Keep default order of required dependencies for explore page table columns

onursumer:
We want the dependencies ordered alphabetically for the Data Standards page, but keep the default order for the Explore page table columns.


inodb (approved):
Thank you! LGTM!

PR #452, opened 2022-11-15
Add title tag

leexgh:
Old
image

New
image


inodb (approved):
Thanks! I like it

inodb (approved):
Thanks! Made a small tweak to text

I wonder also if there are some nextjs ways to set title from different subpages (e.g. data standards page should update the title in some way)

PR #454, opened 2022-11-28
Release 3

### LINKED ISSUE #447: Data related to conditional columns is missing in Explore page tables
We are missing data for certain columns in the Explore page tabs. For example see https://humantumoratlas.org/explore?tab=biospecimen. Also, see https://github.com/ncihtan/htan-portal/issues/295#issuecomment-1116117195

This is because we are not processing the data properly for conditional data fields in https://github.com/ncihtan/htan-portal/blob/28ddab635a00d37dbac6d10132da54051667f147/data/get_syn_data.py

onursumer:
- Make data fetching and processing scripts compatible with the latest synapse files
- Use metadata column header names to properly parse conditional data (Fix #447)


inodb (approved):
Awesome work! Thanks so much @onursumer !

PR #456, opened 2022-12-02
Added schema viz link and link-out icon

linglp:
Related to the Github issue here: https://github.com/Sage-Bionetworks/schematic/issues/1013

### DIFF
--- a/pages/standards.tsx
+++ b/pages/standards.tsx
@@ -115,12 +115,12 @@ const Standards = (data: StandardsProps) => {
                                 </li>
                                 <li>
                                     <Link href="https://sage-bionetworks.github.io/schema_visualization/">
-                                        <span>
+                                        <a>
                                             <FontAwesomeIcon
                                                 icon={faExternalLinkAlt}
                                             />{' '}
                                             HTAN schema visualization
-                                        </span>
+                                        </a>
                                     </Link>
                                 </li>
                             </ul>
--- a/pages/standards.tsx
+++ b/pages/standards.tsx
@@ -115,12 +115,12 @@ const Standards = (data: StandardsProps) => {
                                 </li>
                                 <li>
                                     <Link href="https://sage-bionetworks.github.io/schema_visualization/">
-                                        <a>
+                                        <span>
+                                            HTAN schema visualization{' '}
                                             <FontAwesomeIcon
                                                 icon={faExternalLinkAlt}
-                                            />{' '}
-                                            HTAN schema visualization
-                                        </a>
+                                            />
+                                        </span>
                                     </Link>
                                 </li>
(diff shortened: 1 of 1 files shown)


pages/standards.tsx
```diff
@@ -111,6 +111,15 @@ const Standards = (data: StandardsProps) => {
                                         Imaging
                                     </Link>
                                 </li>
+                                <li>
+                                    <img
+                                        src="https://static.thenounproject.com/png/1167421-200.png"
+                                        width={20}
+                                    />
```
onursumer, line 118: I think this icon should come after the `Link`, not before it. Also we should try using a [font awesome icon](https://fontawesome.com/v4/icon/external-link) instead of an external png reference.

onursumer (changes requested):

pages/standards.tsx
linglp: Thank you for your feedback @onursumer ! Could you check again? I just updated the code based on your suggestion.

pages/standards.tsx
```diff
@@ -111,6 +113,16 @@ const Standards = (data: StandardsProps) => {
                                         Imaging
                                     </Link>
                                 </li>
+                                <li>
+                                    <Link href="https://sage-bionetworks.github.io/schema_visualization/">
+                                        <a>
+                                            <FontAwesomeIcon
+                                                icon={faExternalLinkAlt}
+                                            />{' '}
+                                            HTAN schema visualization
```
onursumer, line 122: Let's swap the order of the text and the icon for consistency. For example see the `Manual` linkout.

htan_manual_linkout

```suggestion
                                            HTAN schema visualization{' '}
                                            <FontAwesomeIcon
                                                icon={faExternalLinkAlt}
                                            />
```

onursumer (changes requested):

pages/standards.tsx
linglp: Thanks @onursumer ! I just updated the code.

pages/standards.tsx
```diff
@@ -111,6 +113,16 @@ const Standards = (data: StandardsProps) => {
                                         Imaging
                                     </Link>
                                 </li>
+                                <li>
+                                    <Link href="https://sage-bionetworks.github.io/schema_visualization/">
+                                        <a>
```
onursumer, line 118: Should we force opening the schema visualizer on a new tab? We may need to do something like this for that: https://stackoverflow.com/a/71029662

linglp: Yeah that makes sense. Thanks @onursumer ! I just made the change.

onursumer (approved):
Looks good to me

PR #464, opened 2022-12-22
Update assets to included fixed HTAPP H&E images

adamjtaylor:
HTAPP H&E thumbnails and minerva stories had a bright pink background due to the PhotometricInterpretation tag not being set to RGB. This is fixed here.

Also add a few WASHU H&E thumbnails that are missing (still more to do!)


inodb (approved):
Thanks!!

PR #380, opened 2022-04-14
Update htan-imaging-assets.json

adamjtaylor:
Update following release 2.1

PR #467, opened 2023-01-11
Add more H&E assets for Duke

adamjtaylor:
Still missing some, but this adds a lot more


adamjtaylor:
These don't seem to show in the preview. I know there was an issue with our previous duke thumbnails that had come out as 1px wide- are we manually filtering them out somewhere?

adamjtaylor:
Looks like we saw that they were causing a problem in https://github.com/ncihtan/htan-portal/pull/380#pullrequestreview-946257535

And manually exclude them here https://github.com/ncihtan/htan-portal/blob/5ed363759ccfdaa300b8ca809c3b9b7d2ab7ae2e/components/FileTable.tsx#L626-L646

adamjtaylor:
This still leaves the single column thumbnails where we don't have new corrected ones yet. To remove manually.

adamjtaylor:
This now shows correctly in the preview

image

Of note is that some of the TMA fields are missing cores.

inodb (approved):
Great! Thank you!

PR #469, opened 2023-01-18
Add minerva thumbnails

adamjtaylor:
Were we have a Minerva story but are missing a thumbnail because of an error in Miniature (typically caused by blank or low contrast channels or a poorly formatted image) we can take one of the top level jpeg tiles as a thumbnail.


inodb (approved):
Thanks so much!

Issue #468, opened 2023-01-12
Refine portal language on data download


adamjtaylor:
We should refine language across the portal including on the [data download](https://data.humantumoratlas.org/data-download) page and the yellow banner to clarify that data available on SB-CGC via CDS with dbGaP authentication is NOT available to download, but rather accessible for cloud workflows and notebooks in SB-CGC.

adamjtaylor:
FYI @inodb, @clarisse-lau and @aclayton555

PR #475, opened 2023-01-27
Change Data Download to Data Access

inodb:
Fix https://github.com/ncihtan/htan-portal/issues/468

Screen Shot 2023-01-30 at 11 16 27 AM


adamjtaylor (changes requested):
Looks good. Maybe lets just clarify that the imaging data on CDS does not require dbGaP authentication.

inodb:
@adamjtaylor good point - fixed

PR #479, opened 2023-02-14
Rename 'Download Source' filter labels to 'Data Access'

adamjtaylor:
Following up from #475 and to fully complete #468, I've taken a shot at renaming the `Download Source` filter to be `Data Access`.

I've left the underlying attribute name as `downloadSource` but updated the relevant `name`, `displayName` and `title` elements that showed up in a full repo search for `Download Source`.


inodb (approved):
Great catch! Thank you!

PR #492, opened 2023-04-03
Tools page minimal

inodb:
Add all analysis tools that were currently under data > analysis tools
to the new tools table designed by @onursumer


alisman:
@onursumer so we are using the same filter/table components to display the tools?  What is the data source of these tools?

onursumer:
> @onursumer so we are using the same filter/table components to display the tools? What is the data source of these tools?

@alisman I needed to do some refactoring to be able use the same filter and table components. The data source is currently this [tools_custom.json](https://github.com/ncihtan/htan-portal/pull/492/files#diff-fcfc27f90a072230276775c656bca6419fede11dde09e79b471f18fd688a76aa) file, but we will probably add more tools soon.

data/tools_custom.json
```diff
@@ -0,0 +1,122 @@
+[
+  {
+    "": "",
+    "Atlas Name": "N/A",
+    "Grant ID": "",
+    "Tool ID": "HTA8_T0",
```
onursumer, line 6: HTA8 is technically MSK, not sure if it's a big deal though.

PR #495, opened 2023-04-03
Allow multiple publication links in a single column

### LINKED ISSUE #494: Allow multiple links in the publication column
For example, cBioPortal has two publications but right now we only show one of them.

image

onursumer:
Fix #494


inodb (approved):
Sweet - thank you!

PR #502, opened 2023-04-07
Add modal to show all valid values

onursumer:
valid_values_modal


inodb (commented):
@onursumer nice! Thanks! I like it!

image

Maybe we can change this to:

"Number of valid options: 232 (Show all)"

Where "Show all" is clickable

PR #512, opened 2023-06-01
Generalize portal for use with other data repositories (e.g. Gray)

alisman:
Remove all WordPress dependency
Remove "HTAN" specific property names from data model


inodb (commented):
@alisman Thanks! Prolly you are already aware, but just fyi, this currently gives an error:

framework-f404b1cdf7903a86.js:1 TypeError: Cannot read properties of undefined (reading 'lead_institutions')

data/processSynapseJSON.ts
```diff
@@ -120,40 +118,65 @@ function addDownloadSourcesInfo(file: BaseSerializableEntity) {
...
-    const flatData = extractEntitiesFromSynapseData(
+    const AtlasMetaMap = _.keyBy(AtlasMetaData, (a) => a.htan_id.toUpperCase());
+    let flatData = extractEntitiesFromSynapseData(
         synapseJson,
         schemaData,
-        WPAtlasMap
+        AtlasMetaMap
     );
+
+    flatData.forEach((d) => {
```
alisman, line 130: @onursumer here is where we strip out HTAN from the source data so it will fit the generalized (d-HTAN-ified) data model.  We need to add properties here.   I did properties one by one just to prove that things were working.  I suppose we could also just strip "HTAN" from all property names.

onursumer: I think this is trickier than I initially thought. Because processed json file is not our only source of truth. We also pull the data model on the fly from https://raw.githubusercontent.com/ncihtan/data-models/, and certain parts of our implementation relies on the fact that the field names in our actual data match the data model.

For example, we autogenerate columns for Biospeciemen table by using the data model https://github.com/ncihtan/htan-portal/blob/fddc6b45fa0211573046d4e273f554438bf67831/components/BiospecimenTable.tsx#L23-L37

Stripping out HTAN from the field name breaks the custom sort functions because the field names in the data do not match the data model anymore.

Maybe instead of modifying the data, we should define a generic data model and force each data portal to provide a mapper/config for those critical fields. something like `uniqueId -> HTANDataFileID`

lib/helpers.ts
```diff
@@ -22,18 +22,18 @@ if (typeof window !== 'undefined') {
...
 
 export interface BaseSerializableEntity {
     // Synapse attribute names
     AJCCPathologicStage: string;
     Biospecimen: string;
     Component: string;
-    HTANParentID: string;
-    HTANBiospecimenID: string;
-    HTANDataFileID: HTANDataFileID; // this is used as the stable UID
+    ParentID: string;
```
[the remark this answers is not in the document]
onursumer: Right, and probably we should define a better generic model with camelCase field names instead of TitleCase.

PR #532, opened 2023-07-24
Update imaging file download instructions

### LINKED ISSUE #529: Suggested manifest.txt file does not work for s5cmd
Current instructions to create a manifest file for `s5cmd` download seem to be incorrect.

broken_manifest

`gs://...` format does not seem to work with the instructions provided at https://learn.canceridc.dev/data/downloading-data. The correct format should be like

`cp s3://public-datasets-idc/d17ebf63-6551-48fe-bd91-8780ece398dd.dcm .`

EDIT INODB: It looks like this was a recent change (i believe before the example docs showed google cloud cli)

onursumer:
Fix #529

dl_imaging


inodb:
Thanks for the quick fix @onursumer! The text at the top should also be changed:

> If you already have the [Google Cloud CLI](https://cloud.google.com/sdk/docs/install) installed you can start from [step 2](https://learn.canceridc.dev/data/downloading-data#step-2-download-the-files-defined-by-the-manifest) in the download instructions and use the below bucket URLs for the manifest.txt file

could be e.g.:

If you already have the [s5cmd](https://github.com/peak/s5cmd#installationl) installed you can start from [step 2](https://learn.canceridc.dev/data/downloading-data#step-2-download-the-files-defined-by-the-manifest) in the download instructions and use the below bucket URLs for the manifest.txt file

fedorov:
Let's hold this until we discuss this comment: https://github.com/ncihtan/htan-portal/issues/529#issuecomment-1650519419 !

inodb:
@onursumer see discussion in #529. Maybe we can update the PR like this:

- [ ] Use new imaging IDC JSON to get updated less verbose bucket URLs. The JSON might need some transformations either on frontend or before as bigquery gives you something like:
    ```
    {
      "collection_id": "htan_hms",
      "ContainerIdentifier": "HTA7_982_1000",
      "s5cmd_manifest_gcp": "s5cmd cp s3://public-datasets-idc/0ab719e1-824e-4fd9-9987-3d0e023a7723/* .",
      "s5cmd_manifest_aws": "s5cmd cp s3://idc-open-data/0ab719e1-824e-4fd9-9987-3d0e023a7723/* ."
    }
    ```
- [ ] Add some text + button to switch between Amazon or Google bucket URLs: 
    image

fedorov:
I would actually save the user from going to IDC docs, and do something similar to what we do in the IDC portal - specify the command line that needs to be executed. The actual command line will depend on whether you are getting files from AWS or GCP, so it is best to pre-generate that one and show to the user. 

image

onursumer:
Updated the instructions to like this:

download_instructions

fedorov:
Looks good to me!

inodb:
Thanks so much for the quick review @fedorov ! Could you confirm the way the URL is constructed for the viewer in the last commit is correct too?

fedorov:
@inodb this looks good! Really happy to see you guys use this functionality and picking up on the power of IDC+SQL! 👍

inodb:
Great! Thanks @fedorov !

inodb (approved):
Thanks so much @onursumer ! This is great!

PR #535, opened 2023-08-03
Improve data processing scripts

### LINKED ISSUE #533: config-htan.yml is not compatible with the latest schematic
Looks like [config-htan.yml](https://github.com/ncihtan/htan-portal/blob/0c55b6b9b9077edaf0cfebc74eedff9e01117663/data/config-htan.yml) is not compatible with the latest schematic [config_example.yml](https://github.com/Sage-Bionetworks/schematic/blob/develop/config_example.yml) anymore. This breaks `get_syn_data.py` script. We need to update `config-htan.yml` to fix this.

We need to be able to properly run `get_syn_data.py` for the new data release.

onursumer:
- Fix the schematic config file wrt [v23.7.1](https://github.com/Sage-Bionetworks/schematic/releases/tag/v23.7.1) (Fix #533)
- Update `get_syn_data.py` to better handle manual includes.
- Update `processSynapseJSON.ts` to fix 10x Visium assay name.
- Add more `scripts` in `package.json`
- Update readme to better indicate what we replace in `syn_data.json`
- Use Duke's first version of imaging manifest: https://www.synapse.org/#!Synapse:syn39271610.1


inodb (approved):
Thanks!

PR #542, opened 2023-08-31
adapt plot to explore page

alisman:
Add new plots tab

TODO:
- [ ] Issue for cases/samples w/o demographics

### DIFF
--- a/lib/helpers.ts
+++ b/lib/helpers.ts
@@ -170,10 +170,10 @@ export function fillInEntities(data: LoadDataResult): Entity[] {
             specimen.ParticipantID || (parentIdMatch ? parentIdMatch[0] : '');
     });
 
-    // get rid of files
-    data.files = data.files.filter((f) => {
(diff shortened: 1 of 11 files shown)


inodb:
Need to make sure file count lines up with humantumoratlas.org:

image
vs
image

lib/helpers.ts
```diff
@@ -172,6 +174,11 @@ export function fillInEntities(data: LoadDataResult): Entity[] {
             specimen.ParticipantID || (parentIdMatch ? parentIdMatch[0] : '');
     });
 
+    // get rid of files
+    data.files = data.files.filter((f) => {
```
alisman, line 178: we can't do this.  we must deal with these as legitimate

next.config.js
```diff
@@ -14,4 +14,7 @@ module.exports = withMDX({
             },
         ];
     },
+    typescript: {
```
alisman, line 17: fix this

alisman:
@onursumer that sounds good. i can't remember though what the result of this problem is in the interface.

Issue #534, opened 2023-08-02
Use BigQuery tables to pull releasable entity lists


clarisse-lau:
Currently the portal script get_syn_data.py takes a number of json files as input to indicate entity release status. I have pulled information from all of the below json files into two BigQuery tables:
`htan-dcc.released.entities` and `htan-dcc.released.metadata` . 

These tables will be used as a single source of truth for file-level release status for other use cases (including loading BQ/ISB-CGC metadata tables, and CDS metadata template generation), and I am hoping the portal script can also be updated to use these tables rather than a growing number of lists. 

The tables are intended to replace the following json files:
- data/release1_include.json
- data/release1_synapse_metadata.json
- data/release2_include.json
- data/release2_synapse_metadata.json
- data/release3_include.json
- data/release3_synapse_metadata.json
- data/dbgap_release_all.json
- data/dbgap_img_release2.json
- and future release#_xx.json files

It also pulls released IDC entities from `data/idc-imaging-assets.json`

Going forward, data release scripts will append synapse IDs to these tables (e.g. adding rows with Data_Release = 'Release 4.0') rather than generating additional release4_include.json, etc files. 

![Screenshot 2023-08-10 at 9 53 03 AM](https://github.com/ncihtan/htan-portal/assets/63608514/86f31b40-7a83-4669-af9a-6735e50090ba)
![Screenshot 2023-08-10 at 9 53 25 AM](https://github.com/ncihtan/htan-portal/assets/63608514/870d9368-bd15-49cc-afb0-75063067d93d)

clarisse-lau:
Below are examples of how the tables can be directly queried using Python. 
I'm happy to help with configuration, etc as needed - authentication instructions can be found on [Google Cloud Docs](https://cloud.google.com/bigquery/docs/authentication)
```
from google.cloud import bigquery
client = bigquery.Client()

include_release_ids = list(client.query("""
   SELECT entityId FROM `htan-dcc.released.entities` 
   WHERE Data_Release IS NOT NULL
""").result().to_dataframe()['entityId'])

# release3_synapse_metadata.json # table also contains manifestVersion
# SELECT manifestId FROM `htan-dcc.released.metadata` 
# WHERE Data_Release = 'Release 3.0'

# dbgap_release_all.json 
# SELECT entityId FROM `htan-dcc.released.entities` 
# WHERE CDS_Release IS NOT NULL
```

onursumer:
[...]

PR #558, opened 2023-10-26
use entities_v4 file to get dbgap release info

onursumer:
- remove `dbgap_release_all.json` and `dbgap_img_release2.json` dependency
- get the information from `entities_v4.csv` instead

Related to #534

### DIFF
--- a/data/processSynapseJSON.ts
+++ b/data/processSynapseJSON.ts
@@ -26,8 +26,6 @@ import {
 } from '../lib/dataSchemaHelpers';
 import fs from 'fs';
 import csvToJson from 'csvtojson';
-import dgbapIds from './dbgap_release_all.json';
-import dbgapImageIds from './dbgap_img_release2.json';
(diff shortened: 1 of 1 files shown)


data/processSynapseJSON.ts
```diff
@@ -67,9 +65,38 @@ function addReleaseInfo(
     }
 }
 
-function addDownloadSourcesInfo(file: BaseSerializableEntity) {
-    const dbgapSynapseSet = new Set(dgbapIds);
-    const dbgapImgSynapseSet = new Set(dbgapImageIds);
+function getDbgapSynapseIds(entitiesById: {
+    [entityId: string]: ReleaseEntity;
+}) {
+    return getReleaseSynapseIds(
```
onursumer, line 71: Not sure if it's critical, but number of unique ids we get from `entities_v4.csv` does not match the number of unique ids in `dbgap_release_all.json`

PR #562, opened 2023-10-31
Release 4.1

onursumer:
Add new data from SRRS and BU


pages/static/data-updates.html
```diff
@@ -4,6 +4,13 @@
 ---
 
 <h1>Data Updates</h1>
+<h2 id="2023-11-??">November ??th, 2023</h2>
```
onursumer, line 7: what should the release date be?

Issue #550, opened 2023-10-04
Duke Data Issues


inodb:
From Marissa: there seem to be participant and biospecimen IDs that are missing from the Duke Atlas in the "Case" and "Biospecimen" tables. For example:
- [ ] When I filter by Atlas "HTAN Duke" and click on the "Atlas" tab, I see there are 1229 biospecimens listed in the table, but the "Biospecimen" tab shows a different value (1139). The 1229 biospecimens matches what I see when I pull down the HTAN Duke biospecimen metadata table using the DCA via synapse.
- [ ] When I filter by Atlas "HTAN Duke" and click on the "Atlas" tab, I also see only 684 cases listed in table and in the "Cases" tab. However, I see 767 participants when I pull down the HTAN Duke participant metadata tables (i.e. Diagnosis) using the DCA via synapse.
- [ ] When I filter by Assay "Bulk RNA-seq" AND Atlas "HTAN Duke" AND Data Access "Synapse", the "Cases" tab shows 461 and "Biospecimen" tab shows 461, but I expect to see 544 cases and 544 biospecimens.

onursumer:
Looks like we have the same issue for `HTAPP`, `Stanford`, and `WUSTL` as well. The number shown in the Atlas tab does not match the number shown in the Biospecimen tab for those atlases.

Atlas tab shows a precomputed value of biospecimens: https://github.com/ncihtan/htan-portal/blob/a8a4349b2b19a68e870767411f3bd2d4623f2a94/data/processSynapseJSON.ts#L232

Biospecimen tab, on the other hand, relies on the number of filtered samples where we further filter out samples by using filtered cases: https://github.com/ncihtan/htan-portal/blob/a8a4349b2b19a68e870767411f3bd2d4623f2a94/lib/filterHelpers.ts#L98-L101

and for the filtered cases we rely on `cases` of a file:
https://github.com/ncihtan/htan-portal/blob/a8a4349b2b19a68e870767411f3bd2d4623f2a94/lib/filterHelpers.ts#L69

I think the problem here is that when we construct the cases for each file, we primarily rely on the `diagnosis`: https://github.com/ncihtan/htan-portal/blob/a8a4349b2b19a68e870767411f3bd2d4623f2a94/lib/helpers.ts#L185-L191

In Duke's case we only have 684 cases in the `diagnosis` file (synapse id: syn39263335). On the other hand, `demographics` file (synapse id: syn39263164) has 767 cases.

[...]

PR #564, opened 2023-11-06
Do not solely rely on diagnosis when generating case data

onursumer:
Partial fix for #550.

It is still possible to get an incorrect number of biospecimens when data is filtered. This is because biospecimen id may not be included in either diagnosis or demographics data files.


onursumer:
This has been implemented as part of #582 (see e4bc0a29a6abcba89f0de0a9d11fdab22615f1ab)

PR #570, opened 2023-12-01
Update CDS imaging release info

onursumer:
old:
image

new:
image


components/HomePage.tsx
```diff
@@ -55,7 +55,7 @@ const HomePage: React.FunctionComponent<IHomePropsProps> = ({
                     }}
                 >
                     <a style={{ color: 'white' }} href="/data-updates">
-                        Data Release V4.1 (Last updated 2023-11-20)
+                        Data Release V4.1 (Last updated 2023-12-01)
```
onursumer, line 58: not sure if we need to update this one

clarisse-lau: Thanks @onursumer! I think we can leave the date as is, as the data isn't new to the portal

clarisse-lau (approved):

Issue #489, opened 2023-03-20
Incorrect rendering of the show less button


onursumer:
Current behavior: `show less` button appears on the first row of all pages.
Expected behavior: `show less` button should appear only on the corresponding cells of the first row of the first page

How to reproduce:
- Go to https://humantumoratlas.org/explore?tab=file
![image](https://user-images.githubusercontent.com/3604198/226474504-0fd3067f-2199-495a-b4c8-489ea73c8395.png)
- Expand the `Organ` and `Diagnosis` columns by pressing the `show more` button on each column
![image](https://user-images.githubusercontent.com/3604198/226474582-6756da84-6f70-4c26-8987-3984483e26cc.png)
- Go to the next page, you will see `show less` button is incorrectly rendered for the first row of the second page
![image](https://user-images.githubusercontent.com/3604198/226474637-f500b619-bde6-4cee-b55a-4a3fedd54e51.png)

onursumer:
Fixed by #602

PR #602, opened 2024-03-12
🐛 Bug Fix: Show Less Button Incorrect Rendering

SAHU-01:
Fix for #489 

Describe changes proposed in this pull request:

- Add a unique cell identifying value for state management, so that **_show more_** and **_show less_** act for that unique cell alone.

Screenshot:

Given:
image

Before:
image

After:
image

Notify reviewers:
@onursumer

### DIFF
--- a/packages/data-portal-commons/src/components/ExpandableText.tsx
+++ b/packages/data-portal-commons/src/components/ExpandableText.tsx
@@ -8,6 +8,7 @@ interface ExpandableTextProps {
     showMoreText?: string;
     showLessText?: string;
     truncateProps?: TruncateProps;
+    value?: string
 }
 
 const Toggle: React.FunctionComponent<{
@@ -25,13 +26,17 @@ export const ExpandableText: React.FunctionComponent<ExpandableTextProps> = (
     props
 ) => {
     // truncate text by default
-    const [truncated, setTruncated] = useState<boolean>(true);
+    const [truncated, setTruncated] = useState<{ [key: string]: boolean }>({});
 
     const showMoreText = props.showMoreText || 'show more';
     const showLessText = props.showLessText || 'show less';
+    const cellKey = `${props.value}`;
 
     const toggleTruncate = () => {
-        setTruncated(!truncated);
+        setTruncated((prevState) => ({
+            ...prevState,
+            [cellKey]: !prevState[cellKey],
+          }));
     };
 
     const showMore = (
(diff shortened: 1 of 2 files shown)


packages/data-portal-commons/src/components/ExpandableText.tsx
```diff
@@ -25,13 +26,17 @@ export const ExpandableText: React.FunctionComponent<ExpandableTextProps> = (
     props
 ) => {
     // truncate text by default
-    const [truncated, setTruncated] = useState<boolean>(true);
+    const [truncated, setTruncated] = useState<{ [key: string]: boolean }>({});
```
onursumer, line 29: It's confusing for this component to keep the state of multiple values when it only receives a single text value.

packages/data-portal-commons/src/components/ExpandableText.tsx
```diff
@@ -25,13 +26,17 @@ export const ExpandableText: React.FunctionComponent<ExpandableTextProps> = (
...
 
     const showMoreText = props.showMoreText || 'show more';
     const showLessText = props.showLessText || 'show less';
+    const cellKey = `${props.value}`;
```
onursumer, line 33: Not sure if we should introduce the cell key inside this component. What is a cell in the context of this component?

packages/data-portal-explore/src/lib/dataTableHelpers.tsx
```diff
@@ -147,8 +147,7 @@ export function truncatedTableCell<T>(cellData: T) {
     const value = _.isFunction(selector)
         ? selector(cellData)
         : cellData[selector as keyof T];
-
     return value ? (
-        <ExpandableText fullText={value} truncateProps={{ lines: 4 }} />
+        <ExpandableText fullText={value} truncateProps={{ lines: 4 }} value={value}/>
```
onursumer, line 151: It's a bit confusing to pass the same value as two different props. We should probably come up with a better solution.

packages/data-portal-commons/src/components/ExpandableText.tsx
SAHU-01: cellKey in our component acts as a unique identifier for each row's text value within the component's state. Since each row's text can be expanded or collapsed independently, the component needs a way to track the expand/collapse state for each row separately.

packages/data-portal-commons/src/components/ExpandableText.tsx
```diff
@@ -40,6 +40,11 @@ export const ExpandableText: React.FunctionComponent<ExpandableTextProps> = (
         </span>
     );
 
+    useEffect(() => {
+            setTruncated(true);
+          }, [props.fullText]);
```
onursumer, line 45: not sure why but the indentation looks incorrect here, can you make sure that the `prettier` hook is running properly?

packages/data-portal-explore/src/lib/dataTableHelpers.tsx
```diff
@@ -147,8 +147,7 @@ export function truncatedTableCell<T>(cellData: T) {
     const value = _.isFunction(selector)
         ? selector(cellData)
         : cellData[selector as keyof T];
-
     return value ? (
-        <ExpandableText fullText={value} truncateProps={{ lines: 4 }} />
+        <ExpandableText fullText={value} truncateProps={{ lines: 4 }}/>
```
onursumer, line 151: we should just revert the changes in this file for simplicity. we don't really change anything but whitespaces.

onursumer (changes requested):
Thanks for fixing this!

Just requested a few minor changes related to code formatting. Other than that it looks good :raised_hands:

packages/data-portal-explore/src/lib/dataTableHelpers.tsx
```diff
@@ -147,7 +147,7 @@ export function truncatedTableCell<T>(cellData: T) {
     const value = _.isFunction(selector)
         ? selector(cellData)
         : cellData[selector as keyof T];
-
+
```
onursumer, line 150: Looks like this file is not completely reverted. You can probably use some git revert / reset command.

onursumer (approved):

PR #613, opened 2024-03-21
Remove age obfuscation for zero values

### LINKED ISSUE #594: Remove age obfuscation for Age at Diagnosis = 0
When masking ages: https://github.com/ncihtan/htan-portal/blob/master/data/get_syn_data.py#L189-L195 ignore `0` values. 

Age at Diagnosis is a required integer element in the HTAN data model. For some cases e.g. precancers, this field is not applicable. Centers use `0` as an opt-out value to get around the requirement. These should remain as `0` (or can be changed to null /NA) rather than converted to age 18. 

Original ticket from user: https://sagebionetworks.jira.com/jira/servicedesk/projects/HTAN/queues/custom/1/HTAN-376

onursumer:
Fix #594


clarisse-lau (approved):
Looks good!

PR #615, opened 2024-03-28
Allow filtering files by Viewers

### LINKED ISSUE #610: Allow filtering the files tab by the view column
It could be better if `viewers` were just another file attribute rather than mapping them in later. We should add the new attribute in the `processSynapseJSON` script so that we can avoid importing mapper json files in the frontend. We can then refactor the `View` column to use the new `viewers` attribute.

onursumer:
Fix #610

image


packages/data-portal-explore/src/components/FileFilterControls.tsx
```diff
@@ -151,6 +152,31 @@ export const FileFilterControls: React.FunctionComponent<IFileFilterControlProps
...
+                                [FileViewerName.minerva]: 'Minerva',
+                                [FileViewerName.ucscXena]: 'USCS Xena',
+                                [FileViewerName.cellxgene]: 'CellxGene',
+                                [FileViewerName.isbcgc]: 'BigQuery',
+
+                                // excluded or aggregated values:
+                                // we are not supposed to see these as filter options
+                                [FileViewerName.autoMinerva]: 'Auto Minerva', // aggregated as Minerva
+                                [FileViewerName.customMinerva]:
+                                    'Custom Minerva', // aggregated as Minerva
```
onursumer, line 170: We have an option to have both these filters (`Auto Minerva` and `Custom Minerva`) in the filter menu. We don't have to aggregate them as `Minerva`.

packages/data-portal-explore/src/components/FileFilterControls.tsx
```diff
@@ -151,6 +152,31 @@ export const FileFilterControls: React.FunctionComponent<IFileFilterControlProps
...
+                                [FileViewerName.cds]: 'CDS', // excluded (this only appears as a download source)
+                                [FileViewerName.idc]: 'IDC', // excluded (we do not show IDC links anymore)
```
onursumer, line 172: we can probably completely drop support for IDC and remove all related files & code

inodb (approved):
Awesome!! Thank you!

Issue #595, opened 2024-02-28
Add more coarse filter for organ type on the explore page


inodb:
<img width="279" alt="image" src="https://github.com/ncihtan/htan-portal/assets/1334004/bce423f4-c0ab-483b-aff5-51eff8a45d05">

Add an additional column in that filter with a more coarse organ type. We can use the same mapping as on the homepage:

<img width="401" alt="image" src="https://github.com/ncihtan/htan-portal/assets/1334004/1a9277e4-f2ae-49b9-bff3-ab881ac43aa6">

PR #617, opened 2024-04-01
add more coarse filter for organ type on the explore page

Nelliney:
Fixing the #595 issue:
1. Created a filter by Organ Type filter on the organ dropdown box on the explore page.
2. Changed the original 'Organ' label in the dropdown box to the 'Organ Details'.

### DIFF
--- a/packages/data-portal-commons/src/lib/getCaseValues.ts
+++ b/packages/data-portal-commons/src/lib/getCaseValues.ts
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import { Entity } from './entity';
-import { getNormalizedOrgan } from '@htan/data-portal-commons';
+import { getNormalizedOrgan } from './entityReportHelpers';
 
 export function getCaseValues(propName: keyof Entity) {
     return (e: Entity) => {
@@ -12,12 +12,10 @@ export function getCaseValues(propName: keyof Entity) {
     };
 }
 
-export function getNormalizedOrganCaseValues() {
-    return (e: Entity) => {
-        if (e.cases) {
-            return _.uniq(e.cases.map((c) => getNormalizedOrgan(c) as string));
-        } else {
-            return [getNormalizedOrgan(e) as string];
-        }
-    };
+export function getNormalizedOrganCaseValues(e: Entity) {
+    if (e.cases) {
+        return _.uniq(e.cases.map((c) => getNormalizedOrgan(c)));
+    } else {
+        return [getNormalizedOrgan(e)];
+    }
 }
--- a/packages/data-portal-utils/src/lib/types.ts
+++ b/packages/data-portal-utils/src/lib/types.ts
@@ -7,7 +7,6 @@ export interface IAttributeInfo<T> {
 
 export enum AttributeNames {
     // Synapse attribute names
-    OrganType = 'OrganType',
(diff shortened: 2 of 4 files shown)


packages/data-portal-commons/src/lib/getCaseValues.ts
```diff
@@ -10,3 +11,13 @@ export function getCaseValues(propName: keyof Entity) {
         }
     };
 }
+
+export function getNormalizedOrganCaseValues() {
+    return (e: Entity) => {
+        if (e.cases) {
+            return _.uniq(e.cases.map((c) => getNormalizedOrgan(c) as string));
+        } else {
+            return [getNormalizedOrgan(e) as string];
```
onursumer, line 20: `getNormalizedOrgan` already returns a string, `as string` typecast is probably redundant.

```suggestion
            return _.uniq(e.cases.map((c) => getNormalizedOrgan(c)));
        } else {
            return [getNormalizedOrgan(e)];
```

packages/data-portal-commons/src/lib/getCaseValues.ts
```diff
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import { Entity } from './entity';
+import { getNormalizedOrgan } from '@htan/data-portal-commons';
```
onursumer, line 3: better to do a relative import here
```suggestion
import { getNormalizedOrgan } from './entityReportHelpers';
```

packages/data-portal-utils/src/lib/types.ts
```diff
@@ -7,6 +7,7 @@ export interface IAttributeInfo<T> {
 
 export enum AttributeNames {
     // Synapse attribute names
+    OrganType = 'OrganType',
```
onursumer, line 10: we should probably move this below the `// Derived or attached in frontend` comment since this is not an actual Synapse attribute name.

onursumer (changes requested):
Thanks for fixing this!

Needs some minor code improvements, looks good otherwise.

packages/data-portal-commons/src/lib/getCaseValues.ts
```diff
@@ -10,3 +11,13 @@ export function getCaseValues(propName: keyof Entity) {
...
+
+export function getNormalizedOrganCaseValues() {
+    return (e: Entity) => {
+        if (e.cases) {
+            return _.uniq(e.cases.map((c) => getNormalizedOrgan(c) as string));
+        } else {
+            return [getNormalizedOrgan(e) as string];
+        }
+    };
+}
```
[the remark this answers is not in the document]
Nelliney: Done.

packages/data-portal-commons/src/lib/getCaseValues.ts
Nelliney: Done.

packages/data-portal-utils/src/lib/types.ts
Nelliney: Done. Also, I have used the lowercase for the first letter for the Attribute Name to be consistent with existing Derived/attached attribute names.

Nelliney:
Thank you for your feedback. Please let me know if there are any other concerns or recommendations.

onursumer (approved):

PR #629, opened 2024-04-29
publications manifest files (CSV and JSON) added to data folder

alexeliotlash:


onursumer:
publication manifest json file has been added as part of #585, we don't need to merge this PR at this point.

PR #631, opened 2024-04-29
Fix an infinite loop issue with findAllAncestors

### LINKED ISSUE #630: Infinite loop issue when finding schema ancestors
This started to happen probably because of the latest changes in the data schema. The ancestor traversal logic is flawed and needs to be fixed regardless.

image

onursumer:
Fix #630


inodb (approved):
LGTM - thank you!

Issue #625, opened 2024-04-16
Add Treatment filter


inodb:
We have more treatment metadata now and it would be nice to add this in as a filter:

- [ ] Design treatment dropdown
- [ ] Implement treatment dropdown

SAHU-01:
When you mentioned a treatment filter were to referring to something like this?
<img width="1511" alt="image" src="https://github.com/ncihtan/htan-portal/assets/71656941/61321426-5786-4b66-8ab8-d14150e55b86">

onursumer:
@SAHU-01 Yes, it should look something like this. We may need to curate/normalize the values though. For example, we we should probably only list `Chemotherapy` as a filter value instead of `Surgery, Chemotherapy` because `Surgery` is already listed as a separate value.

[...]

PR #635, opened 2024-05-08
✨New Feature: Treatment Filter

SAHU-01:
Fixing #625 

Describe changes proposed in this pull request:

- [x]  Design treatment dropdown
- [x] Implement treatment dropdown
- [x]  Normalize Filter Options

Screenshot:

image

Notify reviewers:
@onursumer

### DIFF
--- a/data/processSynapseJSON.ts
+++ b/data/processSynapseJSON.ts
@@ -430,7 +430,10 @@ function processSynapseJSON(
     // split entity into 2 separate entities, Diagnosis and Demographics, if component is 'SRRSClinicalDataTier2'
     flatData = _.flatten(
         flatData.map((d) => {
-            if (d.Component === 'SRRSClinicalDataTier2') {
+            if (
+                d.Component === 'SRRSClinicalDataTier2' ||
+                d.Component === 'Therapy'
+            ) {
                 // TODO figure out where these should go (demographics vs diagnosis)
                 //  we can't show these at this point because these are only defined in SRRSClinicalDataTier2 schema,
                 //  we don't have these in Diagnosis or Demographics schema.
@@ -896,8 +899,10 @@ function extractEntitiesFromSynapseData(
                             const id = attributeToId[column];
 
                             if (id) {
-                                therapyEntity[id.replace(/^bts:/, '')] =
-                                    record.values[i];
+                                // Ensure TypeScript recognizes therapyEntity properly when indexing
+                                (therapyEntity as any)[
+                                    id.replace(/^bts:/, '')
+                                ] = record.values[i];
                             }
                         });
 
@@ -908,8 +913,6 @@ function extractEntitiesFromSynapseData(
                         therapyEntity.Component = 'Therapy';
                         therapyEntity.assayName = 'Therapy';
                         therapyEntity.level = 'Therapy';
-
-                        entities.push(therapyEntity as SerializableEntity);
                     });
                 }
 
--- a/data/processSynapseJSON.ts
+++ b/data/processSynapseJSON.ts
@@ -387,6 +387,22 @@ function processSynapseJSON(
                 accessory.AccessoryAssociatedParentDataFileID;
         });
 
+    const therapyByParticipantID: {
+        [ParticipantID: string]: SerializableEntity[];
+    } = {};
+
+    flatData
+        .filter((obj) => obj.Component === 'Therapy')
+        .forEach((therapyEntity) => {
+            const participantID = therapyEntity.ParticipantID;
+            if (!therapyByParticipantID[participantID]) {
+                therapyByParticipantID[participantID] = [];
+            }
+            therapyByParticipantID[participantID].push(
+                therapyEntity as SerializableEntity
+            );
+        });
+
     const publications: PublicationManifest[] = (flatData.filter(
         (obj) => obj.Component === 'PublicationManifest'
     ) as unknown) as PublicationManifest[];
@@ -533,6 +549,10 @@ function processSynapseJSON(
         (file as SerializableEntity).demographicsIds = (
             parentData?.demographics || []
         ).map((d) => d.ParticipantID);
+        (file as SerializableEntity).therapyIds =
+            therapyByParticipantID[file.ParticipantID]
+                ?.map((t) => t.synapseId)
+                .filter((id): id is string => id !== undefined) || [];
 
         addDownloadSourcesInfo(file, dbgapSynapseSet, dbgapImgSynapseSet);
         addReleaseInfo(file, entitiesById);
@@ -601,6 +621,9 @@ function processSynapseJSON(
         },
         demographicsByParticipantID: demographicsByParticipantID as {
             [ParticipantID: string]: SerializableEntity;
+        },
+        therapyByParticipantID: therapyByParticipantID as {
+            [ParticipantID: string]: SerializableEntity[];
         },
     };
 }
@@ -858,6 +881,36 @@ function extractEntitiesFromSynapseData(
                 // this is a workaround for missing DatasetName for certain schema ids
                 if (synapseRecords.column_order.includes('Dataset Name')) {
                     attributeToId['Dataset Name'] = 'bts:DatasetName';
+                }
+
+                if (synapseRecords.data_schema === 'bts:Therapy') {
+                    synapseRecords.record_list.forEach((record) => {
+                        const therapyEntity: Partial<SerializableEntity> = {
+                            biospecimenIds: [],
+                            diagnosisIds: [],
+                            demographicsIds: [],
+                            therapyIds: [],
+                        };
+
+                        synapseRecords.column_order.forEach((column, i) => {
+                            const id = attributeToId[column];
+
+                            if (id) {
+                                therapyEntity[id.replace(/^bts:/, '')] =
+                                    record.values[i];
+                            }
+                        });
+
+                        therapyEntity.atlasid = atlas.htan_id;
+                        therapyEntity.atlas_name = atlas.htan_name;
+                        therapyEntity.AtlasMeta =
+                            AtlasMetaMap[atlas.htan_id.toUpperCase()];
+                        therapyEntity.Component = 'Therapy';
+                        therapyEntity.assayName = 'Therapy';
+                        therapyEntity.level = 'Therapy';
+
+                        entities.push(therapyEntity as SerializableEntity);
+                    });
                 }
 
                 synapseRecords.record_list.forEach((record) => {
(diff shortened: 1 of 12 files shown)


SAHU-01:
@onursumer I tried modifying the [processSynapseJson.ts](https://github.com/ncihtan/htan-portal/blob/master/data/processSynapseJSON.ts) to include more treatment data but I'm not sure, how to check whether this added the treatment data to processed_syn_data.json.

onursumer:
@SAHU-01 I think a treatment list like this would work, but we may want to change the way we process the `Therapy` data in processSynapseJSON.ts. I'll add another comment/suggestion for that.

> @onursumer I tried modifying the [processSynapseJson.ts](https://github.com/ncihtan/htan-portal/blob/master/data/processSynapseJSON.ts) to include more treatment data but I'm not sure if this approach would work? image

data/processSynapseJSON.ts
```diff
@@ -409,7 +425,10 @@ function processSynapseJSON(
     // split entity into 2 separate entities, Diagnosis and Demographics, if component is 'SRRSClinicalDataTier2'
     flatData = _.flatten(
         flatData.map((d) => {
-            if (d.Component === 'SRRSClinicalDataTier2') {
+            if (
+                d.Component === 'SRRSClinicalDataTier2' ||
+                d.Component === 'Therapy'
```
onursumer, line 430: This case is specific to `SRRSClinicalDataTier2` where we split the data into two separate entities (Diagnosis and Demographics). This will most likely cause some side effects. I think we should handle `Therapy` in the `getSampleAndPatientData` and attach a new `theraphy` object there.

data/processSynapseJSON.ts
```diff
@@ -849,6 +875,36 @@ function extractEntitiesFromSynapseData(
...
+
+                        therapyEntity.atlasid = atlas.htan_id;
+                        therapyEntity.atlas_name = atlas.htan_name;
+                        therapyEntity.AtlasMeta =
+                            AtlasMetaMap[atlas.htan_id.toUpperCase()];
+                        therapyEntity.Component = 'Therapy';
+                        therapyEntity.assayName = 'Therapy';
+                        therapyEntity.level = 'Therapy';
+                    });
+                }
```
onursumer, line 906: We don't probably need to create a new entity, we can handle the therapies as part of `getSampleAndPatientData`. We will need to pass the `therapyByParticipantID` object to that function.

data/processSynapseJSON.ts
```diff
@@ -597,6 +620,9 @@ function processSynapseJSON(
         demographicsByParticipantID: demographicsByParticipantID as {
             [ParticipantID: string]: SerializableEntity;
         },
+        therapyByParticipantID: therapyByParticipantID as {
+            [ParticipantID: string]: SerializableEntity[];
+        },
```
onursumer, line 625: We may not want to attach this to the final json object. If we add the therapy data in `getSampleAndPatientData`, that may be enough for the frontend components to generate list of therapies for filtering purposes.

onursumer (changes requested):
Thanks for the attempt! We should probably change the way we process therapies.

data/processSynapseJSON.ts
onursumer: Let's keep `therapyByParticipantID` attached to the final result for now to keep things consistent with the current implementation. We can remove it later if necessary.

SAHU-01: You are right, on adding the therapy data in `getSampleAndPatientData`,  the frontend components do generate list of therapies for filtering purposes. I'll remove the same from final json object.

packages/data-portal-explore/src/components/FileTable.tsx
```diff
@@ -571,6 +571,17 @@ export class FileTable extends React.Component<IFileTableProps> {
                 wrap: true,
                 sortable: true,
             },
+            {
+                name: 'Treatment',
+                selector: (file: Entity) => {
+                    return _.uniq(
+                        file.diagnosis.map((d) => d.TreatmentType)
```
onursumer, line 578: we should read the value from `therapy` instead of `diagnosis`

```suggestion
                        file.therapy.map((d) => d.TreatmentType)
```

onursumer (changes requested):
This is almost done. Thanks!

I found a minor bug with the new `Treatment` column in the File table. I think once we fix that, we can update the data files and merge the PR. I can handle the data update part since it requires permission to upload data to a cloud server.

packages/data-portal-explore/src/components/FileTable.tsx
SAHU-01: Thank you for pointing it out, that was an oversight from my end.

inodb:
@SAHU-01 great work! Would it be possible to move the Treatment column to the cases tab? We can still keep it as a file attribute, but maybe better not to show by default

SAHU-01:
@onursumer we can display Treatment column in cases tab, I'll have to figure out how to set the `columnVisibility` of `TreatmentType` to true in `EnhancedDataTable`, then `TreatmentType` will be visible in cases tab as a column.

onursumer:
@SAHU-01 I think we can make the new Treatment column visible by default for the Cases tab, and also keep it in the file tab but hide it there by default.

onursumer (approved):

packages/data-portal-explore/src/components/FileTable.tsx
```diff
@@ -866,12 +866,26 @@ export class FileTable extends React.Component<IFileTableProps> {
...
+                                sortable: true,
+                            });
+                        } else {
+                            columns.push({
+                                name: selectorToColumnName(selector),
+                                selector,
+                                wrap: true,
+                                sortable: true,
+                            });
+                        }
```
onursumer, line 888: This is a bit confusing.

Can we actually keep this part as is and add `theraphy` to `excludeFromOtherColumns` instead? Then we can add `Treatment Type` as a default column and we can still hide it by default.

onursumer (changes requested):
One more minor issue to address, otherwise it looks great!

inodb:
@SAHU-01 two minor tweaks that could be nice:

- [ ] Move treatment filter to between Demographics and Assays (as it is a patient-level filter it feels a bit better there)
- [ ] Data Access button can be less wide. The number of filter buttons are starting to reach the width of my laptop, this might help

SAHU-01:
@inodb would this work?
image

packages/data-portal-explore/src/components/FileTable.tsx
SAHU-01: @onursumer I did try that first, but then in the columns, we don't get a `treatment type` option it only displays `Therapy` :
image


which when selected or clicked throws this error:
image

inodb:
@SAHU-01 perfect - thank you!

packages/data-portal-explore/src/components/FileTable.tsx
```diff
@@ -571,6 +571,17 @@ export class FileTable extends React.Component<IFileTableProps> {
...
+            {
+                name: 'Treatment',
+                selector: (file: Entity) => {
+                    return _.uniq(
+                        file.therapy.map((d) => d.TreatmentType)
+                    ).join(', ');
+                },
+                cell: truncatedTableCell,
+                wrap: true,
+                sortable: true,
```
onursumer, line 583:
```suggestion
                sortable: true,
                omit: true
```

packages/data-portal-explore/src/components/FileTable.tsx
```diff
@@ -918,10 +929,17 @@ export class FileTable extends React.Component<IFileTableProps> {
...
-            _.keyBy(this.defaultColumns, (c) => c.name),
-            () => true
+        const columnVisibilityMap = _.mapValues(
+            _.keyBy(
+                this.defaultColumns,
+                (c) => (c as IEnhancedDataTableColumn<Entity>).name
+            ),
+            (c) =>
+                typeof c !== 'number' &&
+                (c as IEnhancedDataTableColumn<Entity>).name !== 'Treatment'
```
onursumer, line 939: Let's try this instead
```suggestion
                !(c as IEnhancedDataTableColumn<Entity>).omit
```

onursumer (changes requested):
Almost there!

packages/data-portal-explore/src/components/FileTable.tsx
```diff
@@ -918,10 +930,17 @@ export class FileTable extends React.Component<IFileTableProps> {
...
-        this.columnVisibility = _.mapValues(
-            _.keyBy(this.defaultColumns, (c) => c.name),
-            () => true
+        const columnVisibilityMap = _.mapValues(
+            _.keyBy(
+                this.defaultColumns,
+                (c) => (c as IEnhancedDataTableColumn<Entity>).name
+            ),
+            (c) =>
+                typeof c !== 'number' &&
```
onursumer, line 939: I think we can remove this `typeof` check
```suggestion

```

packages/data-portal-explore/src/components/CaseTable.tsx
```diff
@@ -16,6 +16,7 @@ import {
     sortByParticipantId,
 } from '../lib/dataTableHelpers';
 import { DataSchemaData, SchemaDataId } from '@htan/data-portal-schema';
+import { truncatedTableCell } from '../lib/dataTableHelpers';
```
onursumer, line 19: looks like we don't use this function in this component anymore, let's remove the `import` statement

packages/data-portal-explore/src/components/FileTable.tsx
SAHU-01: When i remove the `typeof` check it gives the following error:

```
Property 'omit' does not exist on type 'number | IEnhancedDataTableColumn<Entity> | ((...items: IEnhancedDataTableColumn<Entity>[]) => number) | ... 31 more ... | (<A, D extends number = 1>(this: A, depth?: D | undefined) => FlatArray<...>[])'.
  Property 'omit' does not exist on type 'number'.
```

onursumer: Hmm, interesting. Do you see this error in the console or is this an error generated by the IDE? Sometimes IDEs complain about the code but the actual build is successful. I don't get any errors when I build it with `yarn build`, no errors with `yarn dev` either? Which typescript version are you using?

onursumer: You can also try setting the parameter type instead of type casting. Below code also works for me locally.

```JavaScript
const columnVisibilityMap = _.mapValues(
    _.keyBy(
        this.defaultColumns,
        (c: IEnhancedDataTableColumn<Entity>) => c.name
    ),
    (c: IEnhancedDataTableColumn<Entity>) => !c.omit
);
```

SAHU-01: @onursumer I got this error on my IDE. Currently I don't see it anymore, a bit strange, but I'm pushing the code.

onursumer (approved):

PR #640, opened 2024-05-25
Move viewers to header to clean up atlas tab on smaller screens

inodb:
- move viewer icons to header
- remove lead institution column by default to create more space
- add cBioPortal link for Vanderbilt CRC

image

## Follow-up
- can't figure out a way to keep numbers from breaking when screensize is smaller, e.g.
    image
    can maybe be follow-up ticket


packages/data-portal-table/src/lib/helpers.tsx
```diff
@@ -3,9 +3,10 @@ import * as React from 'react';
 export function getColumnKey(col: {
     id?: string | number;
     name: string | number | React.ReactNode;
+    dropdownName?: string;
 }): string {
     // if no id exists, just use name for key
-    return (col.id || col.name || '').toString();
+    return (col.id || col.dropdownName || col.name || '').toString();
```
onursumer, line 9: we should be able to use `id` instead of introducing a new `dropdownName`

inodb: ah gotcha, thank you!

onursumer (approved):
LGTM!

PR #647, opened 2024-06-10
✨New Feature: Data Standards Page Search Filter

SAHU-01:
Improve navigability of the HTAN data standards

Describe Changes:

- [x] Create a Search Filter
- [x] Populate Data to the Search Filter
- [x] Display the Results of the Search Filter 

Screenshots:
image

image

image

Notify reviewers:
@onursumer
@inodb


onursumer:
@SAHU-01 Thanks for working on this!

The functionality looks good, but I think we can improve the implementation and the UI by reusing the existing components and helpers.

First thing is that we can achieve the current functionality within the [DataSchema](https://github.com/ncihtan/htan-portal/blob/7658f207f7ff991d56333964444e513e3986a698/components/DataSchema.tsx#L186) component instead of modifying all the data standard pages (biospecimen, clinical, imaging, etc.) individually. That would simplify the implementation a lot.

Another thing is that we can possibly reuse the filter components in the same way we use in the [Tools Page](https://github.com/ncihtan/htan-portal/blob/29a9668c2a679db0353790bca132fae5a553a7fb/pages/tools.tsx#L52-L64). So, for example, it would be ideal if we can modify the `DataSchema` component in a way like this:

```JavaScript
<>
    <DataStandardFilterControls
        setFilter={setFilter}
        selectedFiltersByGroupName={selectedFiltersByAttrName}
        selectedFilters={selectedFilters}
        entities={props.schemaData}
        groupsByProperty={groupsByProperty}
    />

    <Filter
        setFilter={setFilter}
        selectedFiltersByGroupName={selectedFiltersByAttrName}
        getFilterDisplayName={getToolFilterDisplayName}
    />

    <DataSchemaTable
        schemaData={filteredSchemaData}
        dataSchemaMap={props.dataSchemaMap}
        title="Data Schema:"
    />
</>
```

and in `DataStadardFilterControls` we can simply have a render block this way:

```JavaScript
    return <FilterControls {...filterControlsProps} />
```

This would be similar to the [ToolFilterControls](https://github.com/ncihtan/htan-portal/blob/7658f207f7ff991d56333964444e513e3986a698/components/filter/ToolFilterControls.tsx) implementation where we reuse most of the existing components and helpers.

inodb:
@SAHU-01 Thanks so much! Great start!

I think the current component for the search can be simplified.

The current filtering component is more for when you want to combine several filters:

image

In this case the typical use case would be to search for a single attribute at a time, so a search box like on Tools page might be more intuitive:

image

SAHU-01:
@onursumer and @inodb the way Tools component has a search in the `EnhancedDataTable`, I've added a search filter and search box for all the data standards pages, the user can either filter or search for an attribute and that attribute alone will be visible, I've disabled expansion of filtered attribute as there were errors, based on our need of how we wish to display the results, I'll modify the same.

Next set of todos from my end would be:

- [ ] In `Imaging` data standards page `Imaging Level 3` is displayed twice, need to fix this along with a similar issue in `Sequencing` data standards page.

- [ ] After discussion displaying the results as per need.

- [ ] Improving code quality and modifying code for reusing more existing helpers

onursumer (changes requested):

components/DataSchema.tsx
```diff
@@ -134,63 +143,165 @@ function getColumnDef(dataSchemaMap?: {
...
-    // (exclude Label and Required columns by default)
-    const availableColumns = props.columns || [
-        ColumnName.Attribute,
-        ColumnName.Description,
-    ];
-    // include Valid Values column only if there is data
-    if (
-        !availableColumns.includes(ColumnName.ValidValues) &&
-        hasNonEmptyValidValues(props.schemaData)
-    ) {
```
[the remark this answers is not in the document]
SAHU-01: Yes, there is no need for the same, I was trying something else with columns earlier and I forgot to put back the original logic before pushing the code.

components/DataSchema.tsx
```diff
@@ -134,63 +143,165 @@ function getColumnDef(dataSchemaMap?: {
...
+            ColumnName.Description,
+            ColumnName.ValidValues,
+        ];
+
+        const columnDef = getColumnDef(props.dataSchemaMap);
+        const columns: IDataTableColumn[] = _.uniq(availableColumns).map(
+            (name) => columnDef[name]
+        );
+
+        const filteredSchemaDataById =
```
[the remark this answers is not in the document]
SAHU-01: I've built the search on top of the filter, so when an individual searches for an attribute then that attribute from filters is selected and the result is displayed, thus I kept this logic because for filtering and displaying data, it's required.

components/ConditionalIf.tsx
```diff
@@ -0,0 +1,73 @@
...
+    props
+) => {
+    const [showModal, setShowModal] = useState(false);
+    const onModalClose = () => setShowModal(false);
+    const onClick = () => setShowModal(true);
+
+    const options = props.attributes
+        .filter((attribute) => attribute.length > 0)
+        .map((attribute) => (
+            <div key={attribute}>- {attribute.toLowerCase()}</div>
```
onursumer, line 42: I think it is okay to keep the original casing for Conditional If. We'll need to add an option like this
```suggestion
            <div key={attribute}>- {props.lowerCaseValues ? attribute.toLowerCase(): attribute}</div>
```

components/DataSchema.tsx
```diff
@@ -106,10 +90,86 @@ function getColumnDef(dataSchemaMap?: {
...
+                                    (typeof dep === 'string'
+                                        ? dep
+                                        : dep['@id']) === schemaData.id
+                            );
+                            if (isDependency && value.attribute) {
+                                conditionalIfList.push(value.attribute);
+                            }
+                        }
+                    }
+                }
```
onursumer, line 120: Better to have this logic as a function in `dataSchemaHelpers`. Something like `findConditionalAttributes`.

components/DataSchema.tsx
```diff
@@ -106,10 +90,86 @@ function getColumnDef(dataSchemaMap?: {
...
+                        schemaData.validationRules.some((rule) =>
+                            Array.isArray(rule)
+                        )
+                    ) {
+                        return 'Enum';
+                    }
+                }
+
+                // Default to String if no validation rules or unrecognized type
+                return 'String';
```
onursumer, line 171: This can also be a utility function in `dataShemaHelpers`. Something like `getDataType`. We can also define an enum for this. 

```JavaScript
enum SchemaDataType {
    Integer = "Integer",
    String = "String",
    Enum = "Enum"
}

File: components/ConditionalIf.tsx [hunk 6]
```diff
@@ -0,0 +1,73 @@
...
+        <Modal show={props.show} onHide={props.onClose}>
+            <Modal.Header closeButton>
+                <Modal.Title>{props.attribute} required if</Modal.Title>
+            </Modal.Header>
+            <Modal.Body>{props.options}</Modal.Body>
+        </Modal>
+    );
+};
+
+const ConditionalIfValues: React.FunctionComponent<ConditionalIfValuesProps> = (
```
onursumer, line 32: This is mostly duplicate of [ValidValues](https://github.com/ncihtan/htan-portal/blob/7658f207f7ff991d56333964444e513e3986a698/components/ValidValues.tsx).

Let's generalize and reuse the `ValidValues` component instead. We can name it `TruncatedValuesList` or something like that.

pages/standard/[id].tsx
```diff
@@ -0,0 +1,78 @@
...
+                    </Col>
+                </Row>
+            </Container>
+            <Footer />
+        </>
+    );
+};
+
+export const getServerSideProps: GetServerSideProps = async (context) => {
+    const { id } = context.params as { id: string };
```
Comment (onursumer) [inline pages/standard/[id].tsx:63]: I think currently the id URL is in the form of `.../standard/bts:Therapy`. It would be better if we can remove the initial `bts:` and have URLs like `.../standard/Therapy` instead.

onursumer (changes requested):
Needs some more refactoring, but looks good otherwise. Thanks!

packages/data-portal-table/src/components/DataTableControls.tsx
```diff
@@ -21,6 +21,8 @@ interface IDataTableControlsProps {
     searchBoxPlaceHolder?: string;
     filterText?: string;
     extraControls?: JSX.Element;
+    downloadButtonLabel?: string;
+    showColumnSelect?: boolean;
```
onursumer, line 25: Let's have a `hideColumnSelect` option instead because we show `ColumnSelect` by default

```suggestion
    hideColumnSelect?: boolean;
```

packages/data-portal-schema/src/lib/dataSchemaHelpers.ts
```diff
@@ -576,3 +582,54 @@ function normalizeEntity(
...
     return !entity ? [] : !Array.isArray(entity) ? Array(entity) : entity;
 }
+
+export function findConditionalAttributes(
+    schemaData: DataSchemaData,
+    dataSchemaMap: { [id: string]: DataSchemaData }
+): string[] {
+    const conditionalAttributes: string[] = [];
+
+    for (const [, value] of Object.entries(dataSchemaMap)) {
```
onursumer, line 592: Can't we use `Object.values()` or `_.values()` instead?

```suggestion
    Object.values(dataSchemaMap).forEach(value => {
```

onursumer (changes requested):

components/DataSchema.tsx
```diff
@@ -169,27 +332,286 @@ const DataSchemaTable: React.FunctionComponent<{
...
-                return _.isEmpty(
-                    getDataSchemaDependencies(schema, props.dataSchemaMap)
-                );
-            }}
-            expandableRows={true}
-            expandableRowsComponent={
-                <ExpandableComponent dataSchemaMap={props.dataSchemaMap} />
-            }
+            downloadButtonLabel="Download Data Summary"
+            showColumnSelect={false}
```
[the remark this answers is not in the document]
SAHU-01: I've updated this!

inodb:
Superseded by https://github.com/ncihtan/htan-portal/pull/679

PR #685, opened 2024-10-09
Use grow instead of min width

### LINKED ISSUE #683: Strange Wrapping in Attribute Column
I am looking at: https://data.humantumoratlas.org/standard/biospecimen

And, I searched for "Morphology"

The first column, labeled Attribute has some strange wrapping issue. This occurs regardless of zoom in/out value.

image

onursumer:
Fix #683


leexgh (approved):
LGTM!

PR #686, opened 2024-10-10
Use coarse organ type when counting organs in the explore summary header

### LINKED ISSUE #663: List the simple organs in the counts rather than detailed
Show Organ counts instead of Organ detailed:

image

image

onursumer:
Fix #663

image

The organ count on the explore page is 22 as opposed to 21 shown on the home page because we filter out `Not Reported` (and `unknown`) there. We can do the same thing for explore page with some additional logic.


inodb (approved):
LGTM!

Issue #632, opened 2024-04-30
 Improve the UI for mixed data selections


aclayton555:
This ticket emerges from internal comments on this ticket: https://sagebionetworks.jira.com/browse/HTAN-394

Following our HTAN Data Release 5.1, we now have data types at all levels that may have in different locations. For example there are now Level 2 data that are expected to the access via CDS (e.g. seq, imaging data), and some Level 2 data types (and still L3 and L4 data types) that are expected to be accessed via Synapse (e.g. RPPA).

From Clarisse:

> I think we can refine the logic and move away from explicit level 1/2 vs.. 3/4 behavior. If data has been categorized as Synapse (Open access) per [Update Data Access filtering logic · Issue #620 · ncihtan/htan-portal](https://github.com/ncihtan/htan-portal/issues/620) behavior can be the same for all files regardless of level. 

> So if a user clicks an individual file from the portal, they will be linked out to Synapse, and if they select multiple files to download they’ll be provided with Synapse CLI instructions.

This ticket is to consider potential UI improvements when a user is selecting files for download.

onursumer:
Fixed by #699

PR #699, opened 2024-10-30
Update "Download selected files" modal

adamjtaylor:
- Refines language and layout across the file download modal
- Lists the modes of file access at the top of the header
- Adds instructions and manifest download for using the gen3 client
- Lists Synapse filenames available to access
- Updates the `DataAvailabilityTable` and `FileFilterControls` use the same lock colours


adamjtaylor:
@inodb @onursumer I ended up futzing with this over the weekend but think it is looking good and has both refined language and added functionality. Here are screenshots of what the download files pop-over looks like now

Screenshot 2024-11-03 at 2 23 31 PM

Screenshot 2024-11-03 at 2 23 43 PM

inodb:
@adamjtaylor wow! Looks great! Thanks for working on getting this sorted before the jamboree! Will be very useful!

onursumer:
Fix #632

packages/data-portal-explore/src/components/FileTable.tsx
```diff
@@ -87,17 +90,49 @@ function generateCdsManifestFile(files: Entity[]): string | undefined {
...
+
+    // Return a single command that pulls all objects at once
+    return guids ? `gen3 --endpoint=nci-crdc.datacommons.io/ drs-pull objects ${guids}` : '';
+}
+
+function copyToClipboard(text: string) {
+    navigator.clipboard.writeText(text).then(() => {
+        alert("Copied to clipboard!");
+    });
+}
```
onursumer, line 121: These functions don't seem to be used anywhere. We should probably remove them.

```suggestion
```

inodb (approved):
Thanks so much @adamjtaylor ! Great work!

PR #708, opened 2024-11-04
Add DIWG ancestry calls

### LINKED ISSUE #670: Add DIWG ancestry calls as attributes at Patient Level
- [ ] In ProcessSynapseJSON, add these as patient level attributes
- [ ] On the Portal we can add a new "Ancestry" Filter to the Demographics dropdown:
    image
- [ ] Data should also be downloadable from the Cases Tab here:
    image

[htan_singlecell_ancestry.zip](https://github.com/user-attachments/files/16680860/htan_singlecell_ancestry.zip)

### Notes
We can average the calls per patient:

>Quick question: I noticed each of the results are per sample. It makes sense as that’s how they were obtained, but I’m wondering if it would be easier for reusers of the data if we share the results as a patient level attribute as well. Basically we would make some informed decision on how to assign a patient’s ancestry given the results from one or more samples

>> Yea, it is a good question. Averaging would be fine with me if they are pretty consistent across samples within a patient.

onursumer:
Fix #670

htan_genomic_ancestry

TODO:

- [x] Remove self-reported (ideally we add this in the demographics metadata when missing)
- [x] Call columns "Genomic Ancestry" or something like that
- [x] Better format the numerical values?


onursumer:
Need to reprocess data since we merged release v6.1 (#706)

onursumer:
Reprocessed the data and rebased

PR #714, opened 2024-11-07
Count atlas table viewers dynamically

onursumer:
All counts are dynamically calculated by using the unfiltered files. cBioPortal counts are still hard-coded since we don't have that information as part of the data.

image


inodb:
@onursumer somehow we're missing the BU H&E slide, any clue what happened there?

onursumer:
- There are these 3 files in the cellxgene mapping file which are completely missing in the Files tab. Not sure if any of them is a BU file though. https://github.com/ncihtan/htan-portal/blob/64e30e59aa4aa92f511838aac36ee40194120245/data/cellxgene-mappings.json#L14-L16
- There is only one CHOP file (`syn51301253`) in the BigQuery mapping file: https://github.com/ncihtan/htan-portal/blob/fb8784a2134c6703d764dfd15146a2de22e70828/data/isbcgc-mappings.json#L6
- There is only one MSK UCSC Xena file (`HTA8_4_100001`) in the mapping file: https://github.com/ncihtan/htan-portal/blob/64e30e59aa4aa92f511838aac36ee40194120245/data/ucscxena-mappings.json#L13
- Missing Autominerva files are tricker to manually identify but it's most likely because those files are not in the mapping file https://github.com/ncihtan/htan-portal/blob/bd880195c6a4d43ed9ba5ddb1859c2a56ec8c618/data/htan-imaging-assets.json

inodb:
@onursumer 

> There are these 3 files in the cellxgene mapping file which are completely missing in the Files tab. Not sure if any of them is a BU file though.

Yes one of these is the BU cellxgene instance (you can see the URL by clicking on the production cellxgene link in the explore page): https://cellxgene.cziscience.com/d/BFAA0C46-7E34-4FA9-B08C-6DC6013B735A.cxg/. Don't think this one was ever fully harmonized/curated btw

> There is only one CHOP file (syn51301253) in the BigQuery mapping file

This might be something @PozhidayevaDarya might be able to provide?

> There is only one MSK UCSC Xena file (HTA8_4_100001) in the mapping file:

I see, yeah i also only see Joe Chan's dataset on their website: https://beta.xenabrowser.net/singlecell/?hub=https://previewsinglecell.xenahubs.net:443&defaultTable=htan. So seeing just 1 in your PR might be correct ☑️, maybe it was a manual typo from before

> Missing Autominerva files are tricker to manually identify but it's most likely because those files are not in the mapping file

Ok cool, less worried about those. Your new implementation is prolly correct and there might've been a manual error before

inodb:
@onursumer let's merge this one btw - i think Subhi is looking into BU, so we can prolly fix that later

PR #728, opened 2024-12-05
Update CDS DRS mapping data

### LINKED ISSUE #719: Two sources of truth for whether a file is on CDS (drs_uri & CDS_Release column)
We currently use two approaches to check whether something is on CDS:

- cds_drs mapping file
- CDS_Release column in the entities file

We don't check if there are any discrepancies between these two (theoretically we might list a file as being on CDS, but then don't actually have the proper drs_uri and can't generate the manifest properly)

Note: we're also adding CDS as a "Viewer" in the code which is prolly better to rename (it's a data repository)

Let's fix this issue once we receive the updated mappings from CDS

onursumer:
Fix #719

From: 
image
to:
image

Lots of images were actually not on CDS yet. We had incorrectly assumed several CDS releases were already available in https://github.com/ncihtan/htan-portal/pull/721). Prior to that the dropdown looked like this, so 6155 new files are now available from CDS (added to the release notes).

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,10 @@ that using these commands (requires access to the htan-dcc google project):
 cd data
 bq extract --destination_format CSV released.entities_v6_1 gs://htan-release-files/entities_v6_1.csv
 bq extract --destination_format CSV released.metadata_v6_1 gs://htan-release-files/metadata_v6_1.csv
+bq extract --destination_format NEWLINE_DELIMITED_JSON released.cds_drs_mapping gs://htan-release-files/cds_drs_mapping.json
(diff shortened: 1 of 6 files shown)


public/syn_data.json.gz
onursumer, line 1: we don't need to update this file. original json should still be the same.

onursumer: hmm, actually, the file has changed. @inodb did you run the python script again to update the data?

onursumer: it's also possible that you had the previous version locally and never updated it to the latest

inodb: good catch - fixed this now!

PR #731, opened 2024-12-17
Pull imaging assets from remote mapping file

adamjtaylor:
The htan-imaging-assets mapping file can be automatically generated and stored in a s3 bucket from the s3 inventory. This PR seeks to pull the mapping file directly rather than hosting in this repo as a separate asset

### DIFF
--- a/data/processSynapseJSON.ts
+++ b/data/processSynapseJSON.ts
@@ -56,7 +56,7 @@ import CUSTOM_MINERVA_STORY_MAPPINGS from './minerva-story-mappings.json';
 import IDC_IMAGING_ASSETS from './idc-imaging-assets.json';
 import CDS_ASSETS from './cds_drs_mapping.json';
 
-const AUTOMINERVA_ASSETS_URL = 'https://d3p249wtgzkn5u.cloudfront.net/final-output/htan-imaging-assets-latest.json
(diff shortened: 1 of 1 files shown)


adamjtaylor:
@inodb @onursumer it looks like this is working but please double check.

data/processSynapseJSON.ts
```diff
@@ -51,10 +52,26 @@ import CELLXGENE_MAPPINGS from './cellxgene-mappings.json';
...
+
+async function fetchAutoMinervaAssets(): Promise<AutoMinerva[]> {
+    const response = await fetch(AUTOMINERVA_ASSETS_URL);
+    if (!response.ok) {
+        throw new Error(`Failed to fetch AUTOMINERVA_ASSETS: ${response.statusText}`);
+    }
+    return response.json();
+}
+
+const AUTOMINERVA_ASSETS = await fetchAutoMinervaAssets();
```
onursumer, line 69: This top-level await is causing a compile error for me:

```
TS1378: Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', 'node16', or 'nodenext', and the 'target' option is set to 'es2017' or higher.
```

We should probably move this inside another function. I'll try to fix it.

inodb:
Thanks Adam! Superseded by: https://github.com/ncihtan/htan-portal/pull/733. Note that the change in that PR does require introducing some process to update these mappings every now and then. Will look into some process for that (same applies to the data standards, cellxgene, etc)

PR #733, opened 2024-12-26
Fetch Autominerva imaging assets from a remote URL

onursumer:
This PR also updates the processed JSON file.
(Supersedes #731)

Before:
image

After:
image


packages/data-portal-commons/src/lib/minervaHelpers.ts
```diff
@@ -0,0 +1,26 @@
+import _ from 'lodash';
+import fetch from 'node-fetch';
+import { AutoMinerva } from './entity';
+
+export const DEFAULT_AUTOMINERVA_ASSETS_URL =
+    'https://d3p249wtgzkn5u.cloudfront.net/final-output/htan-imaging-assets-latest.json';
```
inodb, line 6: @onursumer  I think I prefer including the source file that was used to build the portal in the repo but make it clear what URL it was obtained from. In genome-nexus-importer repo we use `Makefile`s for that purpose

inodb (approved):
Love it! Thank you!

PR #751, opened 2025-01-23
Fix journal column casing, move date next to journal

onursumer:
image

### DIFF
--- a/packages/data-portal-explore/src/components/PublicationTable.tsx
+++ b/packages/data-portal-explore/src/components/PublicationTable.tsx
@@ -158,7 +158,31 @@ export const PublicationTable: React.FunctionComponent<IPublicationTableProps> =
             selector: (manifest: PublicationManifest) =>
                 getPublicationJournal(getSummary(manifest), manifest),
             wrap: true,
-            sortable: true
+            sortable: true,
+        },
+        {
+            name: 'Publication Date',
+            selector: (manifest: PublicationManifest) => getDateTime(manifest),
+            cell: (manifest: PublicationManifest) => {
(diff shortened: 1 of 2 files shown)


packages/data-portal-explore/src/components/PublicationTable.tsx
```diff
@@ -156,9 +156,36 @@ export const PublicationTable: React.FunctionComponent<IPublicationTableProps> =
         {
             name: 'Journal',
             selector: (manifest: PublicationManifest) =>
-                getPublicationJournal(getSummary(manifest), manifest),
+                getPublicationJournal(getSummary(manifest), manifest)?.replace(
+                    /\w+/g,
+                    _.capitalize
+                ),
```
inodb, line 162: @onursumer should the capitlization happen at the process JSON step? Just so we don't have to reapply capitalization at every page we show journal name?

inodb: E.g. journal name also shows up here: https://htan-portal-nextjs-git-fork-onursumer-fix-pub-table-bb7c82-htan.vercel.app/publications/hta4_2024_cancer-discovery_liron-d-grossmann?tab=overview

onursumer: Good catch! For now I guess we can move this logic into `getPublicationJournal` because we are already utilizing that function.

It's probably better to normalize data in process JSON step by merging `Publication Manifest` (comes from our internal publications manifest JSON) with `Publication Summary` (comes from an external API). We should probably do that refactoring in a separate PR.

inodb (approved):

PR #752, opened 2025-01-23
Minor performance improvements

onursumer:
- Improve the initial loading of the explore page by loading the schema data as a static prop and fixing redundant rendering
- Simplify the filtering logic of the publications page


onursumer:
superseded by #788

PR #786, opened 2025-03-25
Add How to Cite page

### LINKED ISSUE #766: Create a how to cite page
We don't explicitly mention how to cite the data.

Maybe we do something similar as cBioPortal?

Image

Image

We could e.g. mention to cite the perspective paper, DCC paper and the source article. Maybe a DOI link?

onursumer:
Fix #766

image


inodb:
@onursumer can we link here instead: https://docs.humantumoratlas.org/data_access/citing_htan/

Sorry Jen just set up that page today I think

Issue #821, opened 2025-07-29
Clickhouse Follow Up


inodb:
- [ ] HTAN Parent Data File ID is not showing up in "View Metadata:, see e.g. `MBC_sc_L4.tsv` (but funny enough it's showing up twice in production as `HTAN Parent Data File ID` and as `Parent Data File ID`). I think we might only show associated biospecimen but not associated data files?
- [ ] There are some "useful to code but not useful to end user" fields showing up, e.g. `Is Raw Sequencing`
- [ ] Way less metadata for cases and biospecimen (mostly empty but would be good to keep non-empty somehow)
- [ ] https://github.com/ncihtan/htan-portal/issues/820

Dev mode issues:
- [ ] Atlas table search breaks with an error: Cannot read "AtlasTable.props" in a reactive context, as it isn't observable
- [ ] `Data Standards` and `Tools `page fail loading with a hydration error: Hydration failed because the initial UI does not match what was rendered on the server

PR #824, opened 2025-08-08
Add missing Case and Biospecimen table columns

onursumer:
Related to #821

Adds all the missing columns that are already in the clickhouse DB to the Cases and Biospecimen tables. Those newly added columns are initially hidden. All columns visible or hidden are included in the downloaded metadata file.


inodb (approved):
LGTM - thank you!

Issue #835, opened 2025-09-23
Dev mode bugs post clickhouse


inodb:
There are a few remaining dev bugs after migrating to clickhouse:

- [ ] Atlas table search breaks with an error: Cannot read "AtlasTable.props" in a reactive context, as it isn't observable
- [ ] `Data Standards` and `Tools `page fail loading with a hydration error: Hydration failed because the initial UI does not match what was rendered on the server

PR #849, opened 2025-12-02
Fix atlas table search error in dev mode

onursumer:
Related to #835

(Fixed with help of @claude)

Key Changes:
- Converted getters to methods that accept parameters instead of reading from props
- Moved all data derivation into the render() method where props can be safely accessed
- Pass computed values as parameters to methods that need them

This approach ensures that props are only accessed in the render method (which is safe), and all derived calculations happen with the data passed as parameters rather than reading from the reactive this.props object.


onursumer:
This PR also fixes a bug where the number of total files for each atlas displayed incorrectly when there are filters

Before:
image

After:
image

PR #909, opened 2026-06-29
Fix files wrongly shown as "coming soon" when they are in CRDC

### LINKED ISSUE #908: files sometimes reported as "coming soon" when they are actually in CRDC - display logic bug
When marking files for download, the logic to determine whether each file is available in CRDC or instead is merely "coming soon" seems to be incorrectly tied into the display logic in various ways. For example, starting at this view: (but the problem seems to occur at `/explore` too)

https://humantumoratlas.org/publications/hta7_2022_cell_jia-ren-lin?tab=cycif

If you click the topmost checkbox to select all, the download button changes to "Download 601 selected files" (correct). Upon clicking the download button, the modal with the download instructions shows the Level 3 files as all "Available through Synapse" (correct) but the Level 2 files are all "Coming soon" (wrong). If you close the modal, cause the listing to refresh by e.g. clicking on a table header to sort, then click the Download button again, the modal now shows 50 of the Level 2 files as "Available through NCI CRDC" but the remainder are still "Coming soon" (wrong). Note that the default "Rows per page" is also 50 -- I suspect the download availability logic is only looping over the items currently being displayed, or something along those lines. If you now change the rows per page to 500, it is much more likely that the Download button will correctly list all Level 2 files as "Available through NCI CRDC", but by clicking various sort options and changing the rows per page I can still usually confuse it into showing inaccurate results.

This came up when someone emailed us to ask why none of our data is available through the HTAN portal! It is absolutely a problem for real users, not some obscure corner case.

adamjtaylor:
Fixes #908.

The logic that decides whether each file is available through CRDC or merely "coming soon" was tied into the display logic, exactly as the issue describes.

The download modal categorizes each file by reading `file.viewers?.crdcGc`. That field is only populated as a side effect of `addViewers(file)`, which is called from inside the "View" column's cell renderer in `FileTable.tsx`. react-data-table only renders cells for the rows currently on screen (default 50 per page), so a file only learns that it is in CRDC once its row has actually been rendered. Any file the user has not yet paged or sorted into view still has `viewers.crdcGc` unset, and the modal falls through to the imaging Level 1/2 "coming soon" branch and mislabels it.

This accounts for everything in the report: Level 3 files always showed correctly because their Synapse status is set at ingestion rather than at render time; the Level 2 files flipped to "Available through NCI CRDC" in chunks of ~50 as you sorted because that is one rendered page; and raising "Rows per page" to 500 mostly hid the problem because more rows got rendered. The modal does receive the full selection (the "601 selected files" count is correct) -- the CRDC data just had not been attached to most of those files yet.

## Change

- Add a pure `getCrdcGcAsset(file)` helper in `fileHelpers.ts` that looks the file up in the CRDC/DRS mapping directly, with no mutation and no dependency on whether the row has been rendered.
- Use `getCrdcGcAsset(f)` for the modal's availability categories and for both manifest generators (`crdc_gc_manifest.csv`, `gen3_manifest.json`) instead of the render-populated `viewers.crdcGc` field.
- Leave `addViewers` in the "View" cell, where it is still needed to render the per-row viewer links.

Availability is now derived from the mapping at the moment it is needed, so it no longer depends on pagination, sort order, or rows-per-page.

### DIFF
--- a/packages/data-portal-commons/src/lib/fileHelpers.ts
+++ b/packages/data-portal-commons/src/lib/fileHelpers.ts
@@ -29,6 +29,16 @@ const AUTOMINERVA_MAPPINGS: {
     [synapseId: string]: AutoMinerva;
 } = _.keyBy<AutoMinerva>(AUTOMINERVA_ASSETS, 'synid');
 
+// Pure lookup for a file's CRDC-GC asset. Unlike reading `file.viewers?.crdcGc`,
+// this does not depend on addViewers() having already mutated the file, so it is
+// safe to use for download availability regardless of render/pagination state.
+export function getCrdcGcAsset(
+    file: BaseSerializableEntity,
+    crdcGcMappings: { [fileId: string]: CrdcGcAsset } = CRDCGC_MAPPINGS
+): CrdcGcAsset | undefined {
+    return crdcGcMappings[file.DataFileID];
+}
+
 export function addViewers(
     file: BaseSerializableEntity,
     ucscXenaMappings: { [fileId: string]: string } = UCSCXENA_MAPPINGS,
--- a/packages/data-portal-explore/src/components/FileTable.tsx
+++ b/packages/data-portal-explore/src/components/FileTable.tsx
@@ -42,6 +42,7 @@ import {
     addViewers,
     commonStyles,
     DownloadSourceCategory,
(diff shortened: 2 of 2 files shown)


onursumer (approved):
LGTM!

PR #926, opened 2026-08-31
Cleanup hardcoded uberon codes, and fix a bug with phase2 import script

### LINKED ISSUE #925: Cleanup temporary organ exclusions
Clean up the code related to filtering out hardcoded organ types. This should be fixed already in the BQ database. We don't need to do filtering in the import script.

Related code segment: https://github.com/ncihtan/htan-portal/blob/70dbf07fbb51a290c01f8321457864f1aa24fd9c/build_db/phase2_import.js#L25-L28

onursumer:
Fix #925


inodb (approved):
LGTM

PR #934, opened 2026-09-14
Remove Synapse ids from download metadata popup

### LINKED ISSUE #929: Remove synapse IDs from the metadata download popup
Looks like people are getting confused about the synapse ids listed in the `Download Metadata` popup. (see https://sagebionetworks.jira.com/browse/HTAN-900)

Those files are only downloadable upon clicking on the corresponding synapse ids. For most users, they are not directly accessible on synapse. Let's make it more clear by replacing the synapse ids with something more obvious and simple. 

Image

onursumer:
Fix #929

- Rename Synapse ID column to Download
- Replace synapse ids with a download icon
- Improve column content alignment

Screenshot 2026-09-14 at 2 23 08 PM


adamjtaylor (approved):
LGTM

